PR #27, opened 2018-05-07
Remove 'None' from log level lookups.

febbraiod:
Motivated by a request to not mark log level overrides as 'required' in our Hoist React rest forms, which would mark the field as invalid if null, which it was not.

However, if the comboField was handling a non-required field it would include a null option, redundantly providing both 'None' and 'null' as options in the case of log level overrides. Anselm suggested getting rid of 'None' as a lookup.

Sencha Apps will require a front end change for this to transition smoothly:

https://github.com/exhi/hoist-sencha/pull/54


lbwexler:
This was a nice (minor) cleanup, but was not worth the contortions on the hoist-sencha side.

PR #28, opened 2018-05-07
Do not save null string as message on ClientError, Fixes #21

### LINKED ISSUE #21: Review client error handling
Client error msgs are/were sometimes being saved as the string "null" but should not. Also have seen the server returning messages as  {msg: "[Ljava.lang.String;@e97529"} .

febbraiod:


grails-app/services/io/xh/hoist/clienterror/ClientErrorService.groovy
```diff
@@ -29,7 +29,7 @@ class ClientErrorService extends BaseService implements EventPublisher {
             idSvc = identityService,
             authUsername = idSvc.getAuthUser().username,
             values = [
-                    msg: message,
+                    msg: message == 'null' ? null : message,
```
febbraiod: I have to admit this feels fiddly to me. Null gets converted to a string and we convert it back?

lbwexler:
We cleaned this up on the front end.  I agree with Don -- Fiddly.

PR #31, opened 2018-05-25
Unwind multi-environment config concept

cnrudd:
Still need to test the config differ.  But these changes are pretty definite.  Instructions for migrating to this new scenario are in the ticket.


grails-app/services/io/xh/hoist/config/ConfigService.groovy
```diff
@@ -175,23 +175,12 @@ class ConfigService extends BaseService implements EventPublisher {
         if (valueType != c.valueType) {
             throw new RuntimeException('Unexpected type for key: ' + name)
         }
-        return getInternal(c, notFoundValue)
+        return getInternal(c)
```
lbwexler: Trying to understand how we ever could have returned notFoundValue here before, and why it is ok to remove now

PR #44, opened 2018-06-20
IdentityService.getUser() should not throw when called outside context…

lbwexler:
IdentityService.getUser() should not throw when called without Request


amcclain:
Great thank you for getting this in so quickly.  I will queue up a release to fix my status monitor issues.

PR #55, opened 2018-10-05
Support rendering a cached JSON object at the top-level of a return

amcclain:


amcclain:
An easy way to test this is to add a method to Toolbox `NewsController` -

```
def first() {
  renderJSON(newsService.newsItems.first())
}
```

I just pushed an update to set `cacheJSON` to true on NewsItem.

PR #56, opened 2018-10-12
Checkpoint on identity/auth/roles re-org

amcclain:
+ Identity service caches only username strings in session, delegates to UserService to resolve to HoistUser objects.
+ New requirement for apps to implement a RoleService (over new BaseRoleService), called by HoistUser to produce role assignments on demand and to avoid caching in the user object.
+ HoistUser no longer serializes roles in its JSONFormat to aid more efficient use of this object as a general class for data on enterprise users (which might or might not have app-level roles).
+ Methods added to BaseAuthenticationService to reduce app-level boilerplate, including revised stubs for login/logout + resource whitelisting.
+ Server <> Client API changes around checking for initial auth status.
+ Rename `HoistImplController` -> `XhController` for more consistent use of `XH` as private/implementation namespace, easier to type URLs for debugging.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,4 +1,4 @@
-xhReleaseVersion=5.0-SNAPSHOT
+xhReleaseVersion=6.0-SNAPSHOT
 grailsVersion=3.3.8
 grailsAsyncVersion=3.3.2
 gormVersion=6.1.10.RELEASE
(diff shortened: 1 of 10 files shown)


amcclain:
@lbwexler I'm going to merge this now as discussed so I can release as hoist-core 5.0.0.  I was going to keep building this branch as the snapshot, but alas can't bring snaps through into client site at the moment. And we need to keep moving.

Prior Hoist Core 4.3.0 was released and has all the latest changes prior to this update, so apps that aren't ready to take this yet can stay there w/o missing out on anything else.

Let's review - also need to update the readme!

PR #59, opened 2018-11-03
Improve how Configs and Pref are serialized to client.

lbwexler:
+ REST endpoints should also receive value data in the correct type.
+ DRY


shravnk:
Thanks, I was debating where the boolean handler should go, but this makes it clear that it really belongs on the server. I'm curious about the jsonAsObject option for a user pref - can we now remove a JSON.parse from the client side code?

PR #66, opened 2018-12-18
Additional AppEnvironment enums for UAT+BCP

### LINKED ISSUE #63: UAT AppEnvironment
The standard environment names for at least one of our clients is:
* Dev/Staging (sometimes separate, sometimes synonymous)
* UAT
* Production

A developer at a client was asking how to get UAT to show up in the tab title for a hoist-react app and tried setting the `environment` to UAT in the instance config, but since it is not a recognized values in the enumeration it falls back to Development.

We have a Beta environment, which we could just rename as UAT if we aren't using that elsewhere, or we could just add UAT as a new entry.

Another question I would ask is whether we really need to be so restrictive with these environment names? Seems like we only really use this value to determine if we are in Production or not, otherwise it is just used for display purposes or provided for application-specific logic (though I may have missed something in my searches). Being more flexible could also be useful if there are multiple nodes behind a load balancer, so you could have UAT1, UAT2, etc. which can sometimes be useful for debugging.

amcclain:
+ Fixes #63

### DIFF
--- a/src/main/groovy/io/xh/hoist/AppEnvironment.groovy
+++ b/src/main/groovy/io/xh/hoist/AppEnvironment.groovy
@@ -13,6 +13,14 @@ import io.xh.hoist.json.JSONFormat
 /**
  * Enum describing available application deployment environments.
  *
+ * Hoist itself makes no assumptions or guarantees as to how any given app will behave within a
+ * given environment.
+ *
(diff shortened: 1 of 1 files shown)


amcclain:
@haynesjm42 saw you needed this - see my comments on the linked ticket arguing for this quicker fix to keep moving.

PR #69, opened 2018-12-28
Fix track logging for impersonation

amcclain:
+ Passing incorrect message vs. msg config to trackService silently failed to persist the track log entry.
+ Update TrackService to log an error on write failure within async block.

### DIFF
--- a/grails-app/services/io/xh/hoist/track/TrackService.groovy
+++ b/grails-app/services/io/xh/hoist/track/TrackService.groovy
@@ -45,7 +45,7 @@ class TrackService extends BaseService implements EventPublisher {
         try {
             createTrackLog(params)
         } catch (Exception e) {
-            log.error("Exception writing track log: ${e.message}")
+            logErrorCompact('Exception writing track log', e)
         }
     }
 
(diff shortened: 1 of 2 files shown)


amcclain:
I noticed that the impersonation messages logged to the console were weirdly unhelpful - they didn't include a message to tell you if you were entering vs. exiting impersonation, or indicate who was being impersonated.  Culprit turned out to be using `message` vs. `msg` within map passed to TrackService.

Adjusted w/additional try/catch to avoid silently failing within async block if TrackLog fails to save. Also adjusted error logs within both try/catches to avoid printing the entire stack.

Thought about also supporting `message` as a fallback within TrackService - that's generally our standard - but decided to avoid that messiness. 

Question - for these methods that take a map, is there any way to get more assistance with the supported keys?  Even in the doc string - ironic that our standards on the server are now weaker than they are in the JS client code for documenting public service methods that take a single map arg.  I will do some more research to see if there are any options....

amcclain:
My goodness thank you for the fix there (and then the merge).....

PR #70, opened 2018-12-29
InstanceConfigUtils warns if cannot find/read file

### LINKED ISSUE #58: getInstanceConfig should fail more visibly
Right now, it fails very silently, leaving a mystery as to why app is not finding things.  Easy to screw up if you get a path wrong in yml, etc.

Should throw, or log clearly

amcclain:
+ Fixes #58


amcclain:
Does use of `println` seem reasonable enough here? I think it is worth it to maintain the ability to customize the logger path via instanceConfig - although of course we could use some other approach.  In practice we've done this to write logs to a directory within a docker volume.

lbwexler:
totally reasonable -- and so much better than what we have which is nada!

PR #72, opened 2019-01-16
Log monitor results

shravnk:
- create new appender for MonitoringService only
- log results each time all Monitors are run

### DIFF
--- a/grails-app/services/io/xh/hoist/monitor/MonitoringService.groovy
+++ b/grails-app/services/io/xh/hoist/monitor/MonitoringService.groovy
@@ -87,6 +87,7 @@ class MonitoringService extends BaseService implements AsyncSupport, EventPublis
             markLastStatus(newResults, _results)
             _results = newResults
             evaluateProblems()
+            logResults()
         }
     }
 
@@ -154,6 +155,21 @@ class MonitoringService extends BaseService implements AsyncSupport, EventPublis
     private MonitorStatusReport generateStatusReport() {
         def results = results.values().toList()
         new MonitorStatusReport(results: results)
+    }
+
+    private void logResults() {
+        results.each{
+            def result = it.value,
+                code = result.code,
+                status = result.status.toString(),
+                metric = result.metric ?: null
+            log.info("monitorCode=${code} | status=${status} | metric=${metric}")
+        }
+
+        def failsCount = results.count{it.value.status == FAIL},
+            warnsCount = results.count{it.value.status == WARN},
+            okCount = results.count{it.value.status == OK}
+        log.info("fails=${failsCount} | warns=${warnsCount} | okays=${okCount}")
     }
 
     private void onNotifyTimer() {
--- a/grails-app/services/io/xh/hoist/monitor/MonitoringService.groovy
+++ b/grails-app/services/io/xh/hoist/monitor/MonitoringService.groovy
@@ -158,16 +158,16 @@ class MonitoringService extends BaseService implements AsyncSupport, EventPublis
     }
 
     private void logResults() {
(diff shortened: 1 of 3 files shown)


grails-app/services/io/xh/hoist/monitor/MonitoringService.groovy
```diff
@@ -156,6 +157,21 @@ class MonitoringService extends BaseService implements AsyncSupport, EventPublis
         new MonitorStatusReport(results: results)
     }
 
+    private void logResults() {
+        results.each{
```
amcclain: You could use `results.each{code, result -> ... }` here - this is often handy when iterating over maps in Groovy where you want to get nice variable names for both the key and the value in your closure (vs. getting the default `it` as a map entry).

grails-app/services/io/xh/hoist/monitor/MonitoringService.groovy
```diff
@@ -156,6 +157,21 @@ class MonitoringService extends BaseService implements AsyncSupport, EventPublis
...
+            def result = it.value,
+                code = result.code,
+                status = result.status.toString(),
```
amcclain: You can skip the `toString()` as all you are doing with the status is putting it in a gstring `"${status}" which will call toString() on its own.  (But good to be conscious of this generally, as passing a non-string _directly_ to the logger will throw.)

grails-app/services/io/xh/hoist/monitor/MonitoringService.groovy
```diff
@@ -156,6 +157,21 @@ class MonitoringService extends BaseService implements AsyncSupport, EventPublis
...
+                metric = result.metric ?: null
```
amcclain: Skip this `?: null` here - we actively don't want to take a metric of 0 -> null (always want to be careful with falsey evaluations like this against numbers), and there's no problem with letting null flow into the gstring.

shravnk:
Thank you, made all of the above changes. Definitely agree on the additional monitor config.

Otherwise, day rolling policy works fine in the one example I have. Also apps without any monitors log only the following, which seems reasonable:
`{timestamp} | fails=0 | warns=0 | okays=0`

grails-app/services/io/xh/hoist/monitor/MonitoringService.groovy
```diff
@@ -156,6 +157,21 @@ class MonitoringService extends BaseService implements AsyncSupport, EventPublis
         new MonitorStatusReport(results: results)
     }
 
+    private void logResults() {
```
lbwexler: need space before all of these curly braces where taking closure

src/main/groovy/io/xh/hoist/log/LogUtils.groovy
```diff
@@ -81,10 +82,37 @@ class LogUtils {
...
+            }
+        }
+    }
+
     static void initConfig(Script script) {
         withDelegate(script) {
 
-            def appLogName = Utils.appCode
+            def appLogName = Utils.appCode,
+                    monitorLogName = "${appLogName}-monitor"
```
lbwexler: formattting here.

src/main/groovy/io/xh/hoist/log/LogUtils.groovy
```diff
@@ -94,6 +122,8 @@ class LogUtils {
             }
 
             dailyLog(name: appLogName, script: script)
+            monitorLog(name: monitorLogName, script: script)
+            monitorConsole(name: "monitor-console", script: script)
```
lbwexler: prefer single quotes in groovy (like javaScript)

PR #83, opened 2019-07-10
Add RoutineException and DataNotAvailableException Support

lbwexler:


lbwexler:
All set here for merge.

grails-app/controllers/io/xh/hoist/BaseController.groovy
```diff
@@ -36,15 +36,13 @@ abstract class BaseController implements LogSupport {
...
-    
     //-------------------
     // Implementation
     //-------------------
-    void handleException(Exception ex){
-        if (!ex instanceof NotAuthorizedException) {
-            log.warn(ex.message, ex)  // Problematic requests may or may not be 'errors' or log-worthy
+    void handleException(Exception ex) {
+        if (!(ex instanceof RoutineException)) {
+            logErrorCompact(ex.message ?: 'Exception',  ex)
```
amcclain: I tweaked the doc comment in `RoutineException` to say "Hoist may log exceptions of this type minimally or not at all" as currently I don't think these would make it into the logs in any form.  Is that correct? 

We're also making this change to `logErrorCompact` for all exceptions. I am open to it, but a bit concerned that having the stacktraces for truly unexpected/unhandled exceptions would be useful even if you didn't have the logging in debug mode at the time.

I had been thinking that the new interface would let us continue to log _other_ exceptions with the stacks, but we would aim to be diligent about using `RoutineException` for any exceptions we can anticipate blowing out the logs.

I am not sure - am torn as clearly we can have recurring, frequent _unexpected_ exceptions where we wouldn't use the new type, but also find it frustrating when these make the logs crazy....

(If we do decide to take this route, we could remove the hedging in that doc comment and simply say they won't be logged.)

PR #85, opened 2019-07-23
Pass through export values that have formatting errors

pauldelano:


grails-app/services/io/xh/hoist/export/GridExportImplService.groovy
```diff
@@ -163,17 +163,21 @@ class GridExportImplService extends BaseService {
...
+                        if (metadata.type == 'date') {
+                            value = Date.parse('yyyy-MM-dd', value)
+                        } else if (metadata.type == 'datetime') {
+                            value = Date.parse('yyyy-MM-dd HH:mm:ss', value)
+                        } else if (metadata.type == 'int' || (!metadata.type && value.isInteger())) {
+                            value = value.toInteger()
+                        } else if (metadata.type == 'double' || (!metadata.type && value.isDouble())) {
+                            value = value.toDouble()
+                        }
+                    } catch (Exception ex) {
```
lbwexler: my concern here is that if we log this for every cell, and there are alot of these (say the entire column is "bad" and its a long file), we are going to spam the hell out of the log!

amcclain: Updated to log a count of failures on warn to give some visibility, otherwise trace.

PR #88, opened 2019-07-30
 Dedicated Jackson serializer for GString

### LINKED ISSUE #87: Broken GString JSON serialization via Jackson
Attempting to output e.g. a collection of GStrings via `renderJSON()` and our default Jackson serializer returns e.g.

```
[
  {
    "values": [
      1
    ],
    "strings": [
      "Message #",
      " here"
    ],
    "bytes": "TWVzc2FnZSAjMSBoZXJl",
    "valueCount": 1
  },
  {
    "values": [
      2
    ],
    "strings": [
      "Message #",
      " here"
    ],
    "bytes": "TWVzc2FnZSAjMiBoZXJl",
    "valueCount": 1
  },
  {
    "values": [
      3
    ],
    "strings": [
      "Message #",
      " here"
    ],
    "bytes": "TWVzc2FnZSAjMyBoZXJl",
    "valueCount": 1
  }
]
```

😮 

Looks like we will need to register a custom serializer, unless there is a better default behavior we could be specifying.

amcclain:
+ Fixes #87

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,11 @@
     file to expose an endpoint for connections and wire up a `HoistWebSocketHandler` to relay
     connection events to the new service.
 
+### 🐞 Bug Fixes
+
+* Dedicated Jackson JSON serializer added for Groovy `GString` objects - outputs `toString()` as
+  expected vs. obscure/unwanted object representation (#87).
+
 ## 6.0.2 - 2019-07-24
 
 ### 🐞 Bug Fixes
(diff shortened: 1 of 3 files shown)


amcclain:
Yikes!  Bit by this in a client app that was looking to output a list of error messages that happened to be built as GStrings.

Wanted to check if we should be providing some better default toString() serialization or similar that could have handled this, or if the dedicated serializer is the way to go. Are there other Groovy wrapper classes or similar that could have the same kind of issue?

PR #98, opened 2019-10-30
Accept JSON array or object for JSON prefs (Fixes #97)

### LINKED ISSUE #97: setPrefs error when called with a JSON array value for a JSON preference
The `setPrefs` endpoint will throw if a JSON array is provided for a JSON preference, looks like it is coded to only handle JSONObject types instead of looking for the base JSONElement type after parsing the new preference value.

haynesjm42:
Small fix to look for JSONElement instead of JSONObject here when updating preference values via the `setPrefs` endpoint.

#97


lbwexler:
👍

amcclain:
Hi - looks good - will merge now. 

I assume we need this for a particular client application (in addition to generally supporting JSON as you would expect).  Feel free to give me a shout directly re. any client plugins that also need to get updated and re-released to include.

lbwexler:
agree -- meant to merge this!

PR #99, opened 2020-01-03
Common format strings to mirror what we have in FormatDate.js 

cnrudd:


cnrudd:
Tested here: http://tpcg.io/We4mbANF

In java/groovy, I read there is no symbol to make AM/PM lowercase which the 'a' does in Javascript.

PR #100, opened 2020-01-08
Output all exceptions as JSON + remove obsolete browser lockout 

lbwexler:
+ This allows 404 errors to get treated like other errors.
+ This includes removing obsolete html rendering of exceptions
+ Unrelated cleanups of obsolete browser lockout  code in hoist-core


lbwexler:
Reviewed removing the html rendering with ATM

cr: Peter

PR #101, opened 2020-01-08
Improve formatting of ValidationException

peterjcoffman:
This PR will fix the following two issues:
https://github.com/xh/hoist-core/issues/95
https://github.com/xh/hoist-core/issues/68


src/main/groovy/io/xh/hoist/exception/ExceptionRenderer.groovy
```diff
@@ -24,12 +27,24 @@ import static org.apache.http.HttpStatus.SC_NOT_FOUND
 @Slf4j
 class ExceptionRenderer {
 
+    @Autowired
```
lbwexler: cool use of @Autowired

src/main/groovy/io/xh/hoist/exception/ExceptionRenderer.groovy
```diff
@@ -24,12 +27,24 @@ import static org.apache.http.HttpStatus.SC_NOT_FOUND
...
     void render(Throwable t, HttpServletRequest request, HttpServletResponse response) {
         GrailsUtil.deepSanitize(t)
         response.setStatus(getHttpStatus(t))
+
+        if (t instanceof ValidationException) {
+            def errorMessage = t.errors.allErrors.collect{error ->
+                messageSource.getMessage(error, Locale.US)
+            }.join(',')
+
+            t = new RuntimeException(errorMessage)
```
lbwexler: I changed this to RuntimeException (from throwable), but would it perhaps make more sense to just change the message as formatted in the map returned renderAsJSON?  This would allow the rest of the exception (including its name!) to flow through directly to the client

grails-app/controllers/io/xh/hoist/RestController.groovy
```diff
@@ -30,8 +29,8 @@ abstract class RestController extends BaseController {
             doCreate(obj, data)
             noteChange(obj, 'CREATE')
             renderJSON(success:true, data:obj)
-        } catch (ValidationException ignored) {
-            throw new RuntimeException(errorsString(obj))
+        } catch (ValidationException ex) {
+            throw ex
```
lbwexler: Should we just remove this try/catch block?

lbwexler:
Peter -- one additional thing See the following line in toJSON()

`   isRoutine: t instanceof RoutineException`

Should we check for ValidationException here as well?  Ugh.

grails-app/controllers/io/xh/hoist/RestController.groovy
peterjcoffman: We can certainly remove that one.

We can probably remove the `update` try/catch, too, if we look up the record using `read` rather than `get`.

grails-app/controllers/io/xh/hoist/BaseController.groovy
```diff
@@ -41,6 +45,15 @@ abstract class BaseController implements LogSupport {
     //-------------------
     void handleException(Exception ex) {
         def message = ex.message ?: 'Exception'
+
+        if(ex instanceof ValidationException) {
+            def errorMessage = ex.errors.allErrors.collect{error ->
+                messageSource.getMessage(error, Locale.US)
+            }.join(',')
```
amcclain: Tiny point - suggesting joining with a ` | `  or `. ` so if you have multiple invalid fields the resulting message is as readable as possible.

peterjcoffman: Done.

src/main/groovy/io/xh/hoist/exception/GORMValidationException.groovy
```diff
@@ -0,0 +1,17 @@
...
+ *
+ * Copyright © 2020 Extremely Heavy Industries Inc.
+ */
+
+package io.xh.hoist.exception
+
+/**
+ * Exception for use when user input fails GORM validation
+ */
+class GORMValidationException extends RuntimeException implements RoutineException {
```
lbwexler: + Was thinking that we would want to actually keep the same name -- 'ValidationException'  -- think there may be client code that switches on that name.

+ Was also thinking that this constructor could take the GORM ValidationException and do the string formatting/joining internally.  Seems like that would be more encapsulated

grails-app/controllers/io/xh/hoist/RestController.groovy
peterjcoffman: In fact, I don't believe either try/catch is needed. I do not believe we need an explicit discard() either.

lbwexler: Really?  Sounds like you know GORM/hibernate better than I do -- would love for you to explain that to me.  Am generally cautious about making unneccessary changes to the GORM access in this controller

src/main/groovy/io/xh/hoist/exception/GORMValidationException.groovy
peterjcoffman: >Was also thinking that this constructor could take the GORM ValidationException and do the string formatting/joining internally.  Seems like that would be more encapsulated

I tried to do that and couldn't really figure it out. How would I get access to the MessageSource? A Controller can get it through Spring, but the Exception is not created by Spring so it can't get it through autowiring.

PR #110, opened 2020-02-12
Config canonical form

peterjcoffman:
This change causes all JSON configs to be standardized before saving in the DB.

I am not sure this is the fix we want, but it's worth discussing.

### DIFF
--- a/grails-app/controllers/io/xh/hoist/admin/ConfigAdminController.groovy
+++ b/grails-app/controllers/io/xh/hoist/admin/ConfigAdminController.groovy
@@ -7,12 +7,15 @@
 
 package io.xh.hoist.admin
 
+import groovy.util.logging.Slf4j
 import io.xh.hoist.config.AppConfig
 import io.xh.hoist.RestController
 import io.xh.hoist.security.Access
 import org.grails.web.json.JSONObject
+import io.xh.hoist.json.JSON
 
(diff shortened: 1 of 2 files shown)


grails-app/controllers/io/xh/hoist/admin/ConfigAdminController.groovy
```diff
@@ -26,7 +27,10 @@ class ConfigAdminController extends RestController {
     }
 
     protected void preprocessSubmit(JSONObject submit) {
+        def valueType = AppConfig.get(submit.id)?.valueType ?: submit.valueType
```
lbwexler: Does this re-order keys that would otherwise not be re-ordered?

peterjcoffman: Yes, it does. So this would probably be a no-go.

peterjcoffman: That's annoying. Nowadays the keys in a JS object have a meaningful order, but the JSON spec still says the keys in objects are unordered. So we have inconsistent behavior between JS and JSON.

amcclain:
And closing this out as well - we can leave it to the config admin to determine the order of their keys, and how much they care about the indenting. We got the validation into place, and that's what was really important to us.

PR #111, opened 2020-02-13
Strict json validation

peterjcoffman:
This uses JsonSlurper, which is by default stricter than JSON.parse.

It does not strictly follow the JSON spec, however. I have found that it deviates from the spec by allowing trailing commas, and possibly is lax in other ways.

EDIT: Performance may also be improved, but I haven't benchmarked it and I doubt we care.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## 7.0-SNAPSHOT - under development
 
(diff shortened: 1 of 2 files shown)


amcclain:
I'm assuming we're all good to close this out in favor of the new nano parser.

PR #113, opened 2020-02-14
Switch to using nanoJson

peterjcoffman:
This parser claims to adhere exactly to the spec and it's passed every test case I've thrown at it so far.


lbwexler:
Nice -- thanks for digging so deep on this one.

amcclain:
Great - agreed excellent improvement, thank you!

(Note I fixed up the changelog - it's easy in these merges for the change to end up under the previous release version - and also updated Toolbox to put it back to hoist-core 7.0-SNAPSHOT so we can get the new behavior in use.)

PR #117, opened 2020-03-04
Reversed lines reader

peterjcoffman:
This change solves the problem with excessive memory usage. I have verified that it works correctly on a 10 GiB log file.

I still do not like the fact that we read the entire file to get line counts before we can start reading from the end. On my 10 GiB sample it took 25 seconds to load the log file in the browser, and one CPU core was pinned to 100% for all 25 seconds. However, this is still an improvement over the previous version, which would have died from an OutOfMemoryError.

### DIFF
--- a/grails-app/controllers/io/xh/hoist/admin/LogViewerAdminController.groovy
+++ b/grails-app/controllers/io/xh/hoist/admin/LogViewerAdminController.groovy
@@ -7,6 +7,7 @@
 
 package io.xh.hoist.admin
 
+import groovy.util.logging.Slf4j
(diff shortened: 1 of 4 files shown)


grails-app/controllers/io/xh/hoist/admin/LogViewerAdminController.groovy
```diff
@@ -7,12 +7,15 @@
 
 package io.xh.hoist.admin
 
+import groovy.util.logging.Slf4j
```
lbwexler: remove

PR #118, opened 2020-03-04
Streaming log reader

peterjcoffman:


grails-app/services/io/xh/hoist/log/LogReaderService.groovy
```diff
@@ -0,0 +1,107 @@
...
+    // Implementation
+    //------------------------
+    private List doRead(String filename, Integer startLine, Integer maxLines = 10000, String pattern) {
+        def tail = !startLine || startLine <= 0,
+            ret = new ArrayList(maxLines),
+            file = new File(LogUtils.logRootPath, filename)
+
+        if (!file.exists()) throw new FileNotFoundException()
+
+        long maxEndTime = currentTimeMillis() + configService.getLong('xhLogSearchTimeoutMs', 5000)
```
amcclain: I am a-OK with this, but checking that we didn't intend to put this into the bootstrap config, but wanted to provide the fallback value here instead.  I can see how for certain configs we really don't expect admins to need to think about or change, it's not a bad pattern to have them tucked in here in the code as escape hatches w/o cluttering up the admin console with more xh entries.

Just ensuring that was the idea here - which again, makes sense to me 😄

grails-app/services/io/xh/hoist/log/LogReaderService.groovy
```diff
@@ -0,0 +1,107 @@
...
+
+        Closeable closeable
+        try {
+            if (tail) {
+                ReversedLinesFileReader reader = closeable = new ReversedLinesFileReader(file)
```
amcclain: Well that certainly is a convenient class....

grails-app/services/io/xh/hoist/log/LogReaderService.groovy
```diff
@@ -0,0 +1,107 @@
...
+            }
+
+            return ret
+
+        } finally {
+            if (closeable) closeable.close()
+        }
+    }
+
+    private long getFileLength(File file, long maxEndTime) {
```
amcclain: Just curious - only if you happened to test this and recall - how long does this take for a log file of the size our client was describing (1.5g IIRC)?

grails-app/services/io/xh/hoist/log/LogReaderService.groovy
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    private void throwOnTimeout(long maxEndTime) {
+        if (currentTimeMillis() > maxEndTime) {
+            throw new TimeoutException('Query took too long. Log search aborted.')
+        }
+    }
+}
+
+class TimeoutException extends RuntimeException implements RoutineException  {
```
amcclain: This is a good pattern to keep in mind re. `RoutineException`

grails-app/services/io/xh/hoist/log/LogReaderService.groovy
```diff
@@ -0,0 +1,107 @@
...
+                    def throwAway = reader.readLine();
+                    if (throwAway == null) return []
+                }
+
+                long lineNumber = startLine
+                for (String line = reader.readLine(); line != null && ret.size() < maxLines; line = reader.readLine()) {
+                    throwOnTimeout(maxEndTime)
+                    if (!pattern || line.toLowerCase() =~ pattern.toLowerCase()) {
+                        ret << [lineNumber, line]
+                        lineNumber++
```
amcclain: As above

grails-app/services/io/xh/hoist/log/LogReaderService.groovy
```diff
@@ -0,0 +1,107 @@
...
+        try {
+            if (tail) {
+                ReversedLinesFileReader reader = closeable = new ReversedLinesFileReader(file)
+
+                long lineNumber = getFileLength(file, maxEndTime)
+                for (String line = reader.readLine(); line != null && ret.size() < maxLines; line = reader.readLine()) {
+                    throwOnTimeout(maxEndTime)
+                    if (!pattern || line.toLowerCase() =~ pattern.toLowerCase()) {
+                        ret << [lineNumber, line]
+                        lineNumber--
```
amcclain: Don't we need to decrement `lineNumber` every time so they are accurate when we have a pattern filter?

amcclain:
Screen Shot 2020-03-05 at 6 14 05 AM

amcclain:
Fixed @ 5650d247918643b2fd4770e507f7d5fc21594524 and re-released as 6.4.4 as I know we need the build this morning.

PR #131, opened 2020-04-17
Remove extraneous warning

lbwexler:


amcclain:
Great thanks - this makes sense re. the theme pref.

PR #136, opened 2020-04-22
Add Utils.startupTime

amcclain:
+ Serialize with server environment info to capture server startup time.


lbwexler:
Looks great -- can always later tweak if we want to do it in bootstrap process to have it occur at a well-defined point when server is ready.

Issue #74, opened 2019-02-12
Review row count limit in GridExportImplService re. POI streaming API


amcclain:
If a grid export request is over a set limit - currently 10k rows - GridExportImplService.groovy falls back to the Apache POI "streaming" API to generate such a large Excel sheet:

https://github.com/exhi/hoist-core/blob/0c5e0cc0c685f46efc1ce7c827706e1139bb53cf/grails-app/services/io/xh/hoist/export/GridExportImplService.groovy#L83

https://poi.apache.org/components/spreadsheet/how-to.html#sxssf

This has the side effect of dropping all formatting from the export, including tree formatting, in an unexpected way.  We could consider:

* Moving limit to a constant at the top of the class and documenting
* Moving limit to a required config with a note explaining
* Testing observed memory usage / stability at different rowCount levels, to see if we could up to e.g. 20k or 50k, making it less likely we would hit this limit in practice.

lbwexler:
I think we can safely start by adding a config -- can be a single JSON config for export as a whole as it may need to grow.  Can be used to document/warn about the behavior, including with a client -side toast.    

Could also consider doing this related to the byte size, or at the very least cell count - shouldn't column count enter in this as well?

TomTirapani:
Please see pr here to make it configurable: https://github.com/xh/hoist-core/pull/137

Went with a single config rather than the suggested json config, because I felt that individual configs (we currently only have one) are easier to document with the config note

lbwexler:
Not sure I agree -- think when we have a hyper-specific config, that is
likely to change in the future, would rather nest it in something neat like
'xhExportConfig'.  Keeps things a little more tucked away, and also gives
us more flexibility to play with keys used/needed in future releases
without causing a lot of visible churn.

TomTirapani:
Okay, fair enough :) I've updated the hoist-core pr to use a json `xhExportConfig` config instead.

Also added a small hoist-react pr, which uses this config to manage the toasts we show when exporting: https://github.com/xh/hoist-react/pull/1848

lbwexler:
👍

PR #137, opened 2020-04-22
Add config to adjust when excel export switches to streaming API.

TomTirapani:
See issue: #74
See corresponding hoist-react pr: https://github.com/xh/hoist-react/pull/1848

Went with a single config rather than the suggested json config, because I felt that individual configs (we currently only have one) are easier to document with the config note


lbwexler:
Great change/pickup tom.  Thanks

PR #150, opened 2020-06-15
Add ms to elapsed number in log messages

lbwexler:


febbraiod:
Looks pretty straightforward to me.

PR #151, opened 2020-06-22
+ Support for appTimezone and serverTimezone

lbwexler:


grails-app/services/io/xh/hoist/environment/EnvironmentService.groovy
```diff
@@ -0,0 +1,99 @@
...
+
+
+    //---------------------
+    // Implementation
+    //---------------------
+    private TimeZone calcAppTimezone() {
+        def id = configService.getString('xhAppTimezone', 'GMT'),
+            availableIDs = TimeZone.availableIDs
+        if (!availableIDs.contains(id)) {
+            log.error("App TimeZone '$id' not recognized.  Falling back to GMT.")
```
amcclain, line 90: Would reference the exact config name here.

PR #160, opened 2020-08-11
Debug bulkUpdate method in RestController

febbraiod:


grails-app/controllers/io/xh/hoist/RestController.groovy
```diff
@@ -52,7 +52,7 @@ abstract class RestController extends BaseController {
 
     def bulkUpdate() {
         def ids = params.list('ids'),
-            newParams = JSONParser.parseObject(params.newParams),
+            newParams = JSONParser.parseObject(request.inputStream).newParams,
```
febbraiod, line 55: This may not be the right solutions. It forces a annoying call from our fetchService in HR where we need to have both a params object and a body object. But either way our new JSONParser does not take a grails params object.

PR #161, opened 2020-08-11
Fix bulkUpdate method in RestController

febbraiod:


grails-app/controllers/io/xh/hoist/RestController.groovy
```diff
@@ -52,7 +52,7 @@ abstract class RestController extends BaseController {
 
     def bulkUpdate() {
         def ids = params.list('ids'),
-            newParams = JSONParser.parseObject(params.newParams),
+            newParams = JSONParser.parseObject(request.inputStream).newParams,
```
febbraiod, line 55: This creates an ugly call  on the client. Where we include both a params object and a body object. Lee to review.

grails-app/controllers/io/xh/hoist/RestController.groovy
```diff
@@ -51,8 +51,9 @@ abstract class RestController extends BaseController {
     }
 
     def bulkUpdate() {
-        def ids = params.list('ids'),
-            newParams = JSONParser.parseObject(params.newParams),
+        def body = JSONParser.parseObject(request.inputStream),
+            ids = body.ids,
+            newParams = body.newParams,
```
febbraiod, line 56: Note this follows the pattern of the "update" endpoint above this one.

grails-app/controllers/io/xh/hoist/RestController.groovy
```diff
@@ -51,8 +51,9 @@ abstract class RestController extends BaseController {
     }
 
     def bulkUpdate() {
-        def ids = params.list('ids'),
-            newParams = JSONParser.parseObject(params.newParams),
```
febbraiod, line 55: Make sure we're not using this in Dash.

grails-app/controllers/io/xh/hoist/RestController.groovy
lbwexler: still don't fully understand why we need to change this, and risk breaking client applications.  Seems like the current version is arguably consist with *both* bulkUpdate and *delete*

febbraiod: Well I'm pretty sure if a client app is using this, it's already broken.  JSONParser.parseObject(params.newParams) throws a MissingMethodException because params.newParams is a GrailsParameterMap which is in invalid argument for JSONParser.parseObject. "No signature of method: static io.xh.hoist.json.JSONParser.parseObject() is applicable for argument types: (grails.web.servlet.mvc.GrailsParameterMap) values: [[groupName:TEST]]"

We can continue to use the params object by changing to `newParams = params.newParams` if that's preferable. But, the reason I changed to what I did is because the create and update endpoints above this one do not use the params object, but rather `JSONParser.parseObject(request.inputStream)`. So my change produces symmetry with those two endpoints and in the client code where the add and edit actions use a "body" property to pass these params to the fetchService vs. a "params" property:

`const fetchMethod = isAdd ? 'postJson' : 'putJson',
            response = await XH.fetchService[fetchMethod]({url, body: {data}}),`

lbwexler: sounds good

PR #165, opened 2020-09-11
Add JsonBlob domain class, service and controllers

TomTirapani:
See hoist-react PR: https://github.com/xh/hoist-react/pull/2067

### DIFF
--- a/grails-app/controllers/io/xh/hoist/admin/JsonBlobAdminController.groovy
+++ b/grails-app/controllers/io/xh/hoist/admin/JsonBlobAdminController.groovy
@@ -18,7 +18,14 @@ class JsonBlobAdminController extends RestController {
 
     def lookupData() {
         renderJSON(
-            types: JsonBlob.list().collect{it.type}.unique().sort()
+            types: JsonBlob.createCriteria().list{
+                projections { distinct('type') }
+            }.sort()
         )
+    }
+
+    protected void preprocessSubmit(Map submit) {
+        submit.lastUpdatedBy = username
+        if (submit.value) submit.valueLastUpdated = new Date()
     }
 }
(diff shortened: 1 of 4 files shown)


grails-app/controllers/io/xh/hoist/admin/JsonBlobAdminController.groovy
```diff
@@ -0,0 +1,24 @@
...
+import io.xh.hoist.jsonblob.JsonBlob
+import io.xh.hoist.RestController
+import io.xh.hoist.security.Access
+
+@Access(['HOIST_ADMIN'])
+class JsonBlobAdminController extends RestController {
+    static restTarget = JsonBlob
+    static trackChanges = true
+
+    def lookupData() {
```
lbwexler, line 19: Something like the following (from the tracking controller) might be more efficient.  Would be worth maybe seeing the SQL that is produced.

 return TrackLog.createCriteria().list {
            projections { distinct(property) }
        }.sort()

grails-app/domain/io/xh/hoist/jsonblob/JsonBlob.groovy
```diff
@@ -0,0 +1,49 @@
...
+import io.xh.hoist.json.JSONFormat
+import io.xh.hoist.util.Utils
+
+class JsonBlob implements JSONFormat {
+
+    String type
+    String username
+    String name
+    String value
+    String description
```
lbwexler, line 18: Do we want a lastUpdatedBy?   Is there also a way to ensure that valueLastUpdated gets updated when changing in RestEditor?  Might want to try to build it in to the Domain object via a trigger, or else need to add something in the RestController.

I could see adding an additional json blob with arbitrary metadata that is seperate from the value.

PR #169, opened 2020-09-24
JsonBlob ACL

TomTirapani:
+ Implement ACL in JsonBlobService
+ Remove JsonBlob.valueLastUpdated

See hoist-react PR: https://github.com/xh/hoist-react/pull/2092

### DIFF
--- a/grails-app/services/io/xh/hoist/jsonblob/JsonBlobService.groovy
+++ b/grails-app/services/io/xh/hoist/jsonblob/JsonBlobService.groovy
@@ -7,6 +7,7 @@
 package io.xh.hoist.jsonblob
 
 import io.xh.hoist.BaseService
+import io.xh.hoist.exception.NotAuthorizedException
 import io.xh.hoist.json.JSONParser
 
 class JsonBlobService extends BaseService {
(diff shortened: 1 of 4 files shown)


amcclain (approved):
Looks like the simple implementation we discussed.  Hopefully this makes sense generally as a way to reserve the ACL for future sharing capabilities / controls while keeping things fast and simple for now.

grails-app/services/io/xh/hoist/jsonblob/JsonBlobService.groovy
```diff
@@ -15,34 +16,33 @@ class JsonBlobService extends BaseService {
...
         return formatForClient(blob, true)
     }
 
     Map update(int id, String name, String value, String description) {
         JsonBlob blob = JsonBlob.get(id)
+        ensurePassesAcl(blob)
 
         if (name) blob.name = name
+        if (value) blob.value = value
         if (description) blob.description = description
```
lbwexler, line 45: should these be checks for null, not truthy?  Seems like you might want to allow the client to remove the description.  Also, if the client tried to blank the name or value, would like them to get the appropriate validation error,  not just have it be ignored.

lbwexler:
Will merge, deal with my nit in the update method on develop.  Needed for hoist-react which was merged!

PR #171, opened 2020-09-30
Add UUID column to JsonBlobs

TomTirapani:
See hoist-react PR: https://github.com/xh/hoist-react/pull/2102

### DIFF
--- a/grails-app/domain/io/xh/hoist/jsonblob/JsonBlob.groovy
+++ b/grails-app/domain/io/xh/hoist/jsonblob/JsonBlob.groovy
@@ -11,7 +11,7 @@ import io.xh.hoist.util.Utils
 
 class JsonBlob implements JSONFormat {
 
-    String token
+    String token = UUID.randomUUID().toString().substring(0, 8)
     String type
     String owner
     String acl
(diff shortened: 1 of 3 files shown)


grails-app/domain/io/xh/hoist/jsonblob/JsonBlob.groovy
```diff
@@ -40,7 +42,7 @@ class JsonBlob implements JSONFormat {
     }
 
     Map formatForJSON() {[
-        id: id,
+        id: uuid,
```
haynesjm42, line 45: Not sure we want to conflate the id/uuid here. I think we are fine to still return the id just as we do for other domain objects along with the new uuid

lbwexler: I agree with this -- thought we would call the new thing token?

TomTirapani: The name 'token' came later, as a result of john's review. I'm in the process of updating this branch

grails-app/domain/io/xh/hoist/jsonblob/JsonBlob.groovy
```diff
@@ -42,7 +42,8 @@ class JsonBlob implements JSONFormat {
     }
```
lbwexler: no need/value to add a unique constraint to token?

PR #179, opened 2020-10-27
Add JsonBlob meta column

TomTirapani:
See hoist-react PR: https://github.com/xh/hoist-react/pull/2164

### DIFF
--- a/grails-app/services/io/xh/hoist/jsonblob/JsonBlobService.groovy
+++ b/grails-app/services/io/xh/hoist/jsonblob/JsonBlobService.groovy
@@ -25,11 +25,12 @@ class JsonBlobService extends BaseService {
         }
     }
 
-    Map create(String type, String name, String value, String description) {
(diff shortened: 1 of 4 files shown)


grails-app/services/io/xh/hoist/jsonblob/JsonBlobService.groovy
```diff
@@ -90,7 +91,8 @@ class JsonBlobService extends BaseService {
             description: blob.description,
             dateCreated: blob.dateCreated,
             lastUpdated: blob.lastUpdated,
-            lastUpdatedBy: blob.lastUpdatedBy
+            lastUpdatedBy: blob.lastUpdatedBy,
+            meta: JSONParser.parseObjectOrArray(blob.meta)
```
lbwexler, line 95: This seems a bit weird to me that meta would not always be an object.  But I suppose it is consistent with the handling of 'value'.

PR #183, opened 2020-11-30
Add simple MemoryMonitoringService

amcclain:
+ Takes and stores a limited history of memory usage snapshots, based on JVM runtime stats.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
   runtime in the Hoist Admin Console.
 * HTML-encode certain user-provided params to XhController endpoints (e.g. track, clientErrors,
   feedback) to sanitize before storing / emailing.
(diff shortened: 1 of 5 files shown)


amcclain:
See also https://github.com/xh/hoist-react/pull/2212

lbwexler:
Adding a snapshot is an order N operation (due to the need to find the minimum in the  delete).  But think that's fine for the number of snapshots that we have hard-coded that we are going to save.  

Could be a problem if you wanted to do it much more frequently, and we would want a different data structure, but does not seem worth futzing with it

amcclain:
Great - thanks for the careful look. 

I thought about making the max snapshot number configurable, but I didn't want to add another top-level key, and also wanted to use the simple built-in Timer feature to take a config key name and have it auto-adjust its interval on the fly.

I figure we can see what real world use we get out of the feature and evolve from there as helpful.

PR #185, opened 2020-12-01
Add owner and acl to create and update method parameters

petradish:
This PR closes https://github.com/xh/hoist-core/issues/184


lbwexler:
it seems to me that this change by itself is going to break the methods in the XH controller.

We could modify those for security purposes to *never* change the user or acl, but at the very least they need to pass those arguments.

Alternatively, we could add a new server method `updateAccess()` that *only* optionally  changed the owner and the acl.  Might be safer and more straighforward.

PR #186, opened 2020-12-17
Logging fixes

lbwexler:


src/main/groovy/io/xh/hoist/exception/ExceptionRenderer.groovy
```diff
@@ -58,7 +58,13 @@ class ExceptionRenderer {
     // Template methods.  For application override
     //---------------------------------------------
     String summaryTextForThrowable(Throwable e) {
-        return e.message ?: e.cause?.message ?: e.class.name
+        def cause = e.cause,
+            message = e.message,
+            className = e.class.name
+
+        if (message) return "$message [$className]";
+        if (cause?.message) return "${cause.message} [${cause.class.name}]"
```
haynesjm42, line 66: Would we not be interested in the exception class name in addition to the cause message/class name here?

CHANGELOG.md
```diff
@@ -1,5 +1,18 @@
 # Changelog
 
+## 9.0.0 - 2020-12-17
+
+### 🎁 New Features
```
amcclain, line 5: This should be Breaking Changes, with the boom icon

CHANGELOG.md
```diff
@@ -1,5 +1,18 @@
...
+
+* `LogSupport` API enhancements:
+  * `logErrorCompact()` and `logDebugCompact()` now only shows stacktraces on `TRACE`
+  * `withInfo()` and  `withDebug()` now will automatically print "pre-execution" line 
+    if logging level on 'TRACE'.
+  *  `withShortInfo()` and `withShortDebug()` are now obsolete and have been removed.
```
amcclain, line 11:
```suggestion
  *  `withInfo()` and  `withDebug()` now log only once _after_ execution has completed. Raising the log level of the relevant class or package to `TRACE` will cause these utils to also log a line _before_ execution, as they did before. (As always, log levels can be adjusted dynamically at runtime via the Admin Console.)
  * The upgrade to these two utils mean that they **completely replace** `withShortInfo()` and `withShortDebug()`, which have both been **removed** as part of this change. 
```

haynesjm42 (approved):

amcclain (approved):
Took a quick look - looks good - didn't want to dupe too much review work.

Do have some quick suggestions re. changelog

PR #187, opened 2020-12-20
Enhancements to Logging and Exception Logging

lbwexler:
One thing I wasn't *quite* sure about was the format of the username, and whether we want to tag it with a `user:` prefix.  I think we should stay out of the business of tagging right now, but happy to discuss.

Also note the removal of the "[log on trace for more...]".  Happy to discuss that of course but would like to propose dropping that.  

Think its a pretty simple story we have to answer the question "how do I get more information about exceptions in the log"


lbwexler:
cr: ATM

PR #190, opened 2020-12-31
Fix to impersonation tracking

lbwexler:
+ Track start and stop within impersonation session for clarity
+ Use fixed messages with data field for better analysis
+ misc. cleanups to use new username getter
+ be sure to always log ending of impersonation session when moving from user to user.
+ log events to main log/standard out

### DIFF
--- a/grails-app/services/io/xh/hoist/user/IdentityService.groovy
+++ b/grails-app/services/io/xh/hoist/user/IdentityService.groovy
@@ -112,14 +112,20 @@ class IdentityService extends BaseService {
             throw new RuntimeException('Cannot impersonate when outside the context of a request')
         }
         if (!authUser.isHoistAdmin) {
-            throw new RuntimeException("User '${authUser.username}' does not have permissions to impersonate")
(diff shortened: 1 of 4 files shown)


grails-app/services/io/xh/hoist/user/IdentityService.groovy
```diff
@@ -112,14 +112,20 @@ class IdentityService extends BaseService {
...
         if (!authUser.isHoistAdmin) {
-            throw new RuntimeException("User '${authUser.username}' does not have permissions to impersonate")
+            throw new RuntimeException("User '$authUser.username' does not have permissions to impersonate")
         }
         if (!targetUser?.active) {
             throw new RuntimeException("Cannot impersonate '$username' - no active user found")
         }
 
-        trackImpersonate("User '${authUser.username}' is now impersonating user '${targetUser.username}'")
+        // first explicitly end any existing impersonation session -- important for tracking.
```
amcclain, line 121: 👍 I see, good catch

amcclain (approved):
Nice cleanup!

PR #191, opened 2021-03-25
Serialize NaN and Infinity to `null`

lbwexler:
This is consistent with how we serialized Floats and Doubles pre-jackson, and consistent with how the react client expects to receive these numbers.


amcclain:
Looks good 👍

PR #194, opened 2021-04-10
Fix string format constants on DateTimeUtils

amcclain:
+ Incorrect use of D (day of year) vs d (day of month).


lbwexler:
Great Catch!

PR #195, opened 2021-04-10
Respect failed/inactive status set by monitors

amcclain:
+ Status check implementations should be able to mark a check as failed or inactive internally - hopefully installing a message to explain - without the subsequent metric-based evaluation overwriting such a status and message.


amcclain:
Use case / prompted by: https://github.com/xh/toolbox/commit/b5907d26db0f1207f553dba42c4a51e100cfcf68

PR #201, opened 2021-07-22
Unwind removing use of streamingCellThreshold to downgrade excelTable

petradish:
Discussed w/Lee yesterday - will revisit how to make tree grid exports more performant. In the meantime, we do need the streamingCellThreshold to downgrade large exports (realized the limitation in performance was not style-related, though the change to cache groupedColors prevents the apache error of exceeding 64k styles)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,9 +4,8 @@
 
 ### 🎁 New Features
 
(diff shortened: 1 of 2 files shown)


lbwexler:
Just restoring the status quo API, as we are in the midst of several investigations about performance.

Taking the more efficient handling of styling, which is a win.

PR #203, opened 2021-08-02
Include full URL with client error reports

saba-mo:
Should be merged with matching PR in hoist-react: https://github.com/xh/hoist-react/pull/2570


grails-app/domain/io/xh/hoist/clienterror/ClientError.groovy
```diff
@@ -29,6 +30,7 @@ class ClientError implements JSONFormat {
         cache true
         error type: 'text'
         msg type: 'text'
+        url type: 'text'
```
amcclain, line 33: We use the `text` database type for strings that could be arbitrarily long - e.g. JSON blobs like the `error` property here. I am not an expert here, but my general understanding is that these types of columns are less efficient for the database to store vs. the standard `varchar(255)` type that would be used by default.

I'm pretty confident that most if not all URLs we would have in our apps would fit within that default - the only thing I can imagine overflowing it would be a long query parameter or route, and I don't think we have a strong need (or desire) to capture those. It would be worth posting an artificially long URL to ensure it truncates cleanly w/o throwing an error - I think it will.  (Note you will need to drop the column in your local DB to have it re-added as a varchar before testing.)

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 * Excel cell styles with grouped colors are now cached for re-use, avoiding previously common file error that limits 
   Excel tables to 64,000 total styles.
+* Client error reports now include the full URL for additional troubleshooting context.
```
amcclain, line 9: We have a mixed story at the moment when it comes to database modifications required by a change like this. In some client environments, we are OK leaving the Grails `dbCreate` setting on `update` and having GORM detect and make the required schema changes on its own, but support for that in practice is a mixed bag, and in some client DB environments our runtime user doesn't have DBO rights in production to modify the database.

Given that, it's a good practice to call out the fact that a DB update is required - something like:

  * ⚠ NOTE - this requires a new, nullable varchar(255) column be added to the `xh_client_error` table in your app's configuration database. Review and run the following SQL, or an equivalent suitable for the particular database you are using:
```
ALTER TABLE `xh_client_error` ADD COLUMN `url` VARCHAR(255) NULL;
```

amcclain (commented):
Looks great, thanks - I was just wishing for this earlier this morning with a real world error report.  Two suggestions below - will look at the Hoist side now.

PR #210, opened 2021-10-01
Block sending unfiltered email from local dev

amcclain:
+ When running in local development mode, EmailService now requires either a filter or override config to be active, to ensure that a dev has placed some constraints on where dev emails will go.

### DIFF
--- a/grails-app/services/io/xh/hoist/email/EmailService.groovy
+++ b/grails-app/services/io/xh/hoist/email/EmailService.groovy
@@ -49,12 +49,20 @@ class EmailService extends BaseService {
             throwError = args.throwError
 
         try {
+            def override = parseMailConfig('xhEmailOverride'),
+                filter = parseMailConfig('xhEmailFilter')
+
             def originalTo = args.to,
                 sender = args.from ? formatAddresses(args.from)[0] : parseMailConfig('xhEmailDefaultSender')[0],
                 doLog = args.containsKey('doLog') ? args.doLog : true,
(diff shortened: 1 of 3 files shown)


grails-app/services/io/xh/hoist/email/EmailService.groovy
```diff
@@ -64,15 +72,14 @@ class EmailService extends BaseService {
...
             }
 
-            def overrideEmail = parseMailConfig('xhEmailOverride')
-            if (overrideEmail) {
-                toRecipients = overrideEmail
+            if (override) {
+                toRecipients = override
                 ccRecipients = null
-                subj += " [for $originalTo]"
+                subj += " (for $originalTo)"
```
amcclain, line 82: tweak - `originalTo` is a list, so it renders in [] - they looked a bit crazed doubled-up

src/main/groovy/io/xh/hoist/util/Utils.groovy
```diff
@@ -69,6 +70,13 @@ class Utils {
         return appEnvironment == AppEnvironment.PRODUCTION
     }
 
+    /**
+     * Indicates if app is running in local development mode, regardless of AppEnvironment.
+     */
+    static Boolean getIsLocalDevelopment() {
+        return Environment.isDevelopmentMode()
```
amcclain, line 77: I was a bit surprised we didn't have this. I think it's worth it to get "local" in here given possible confusion between Grails and Hoist AppEnvironment.

PR #221, opened 2021-11-15
Log changes

lbwexler:
Note that I left the removal of the static versions of the methods out of the breaking changes -- I don't think there is any real usage there.  They have long caused problems to the IDE and JVM in recognizing method calls on the core methods. 

Please see also:
https://github.com/xh/toolbox/pull/539


amcclain (approved):
This looks great and is a huge cleanup - happy to have this merged in if you are ready for us to do so.  Thanks!

PR #225, opened 2021-12-06
Adjust CHANGELOG re: JsonBlob database migration

TomTirapani:
Paul noticed that our CHANGELOG between 8.5.0 > 8.6.1 requires database changes to the JsonBlob table, but the instructions are incomplete and inconsistent with the released state of the versions


amcclain:
Thanks for taking the time to correct this!

PR #231, opened 2022-01-15
Throw if bootRun task run against plugin

amcclain:
+ Avoid confusion with local development in "wrapper" mode, with both hoist-core and an app setup in a multi-project gradle build.
+ Devs should ensure they run `bootRun` from their app-specific gradle scope, not the combined top-level scope created by the wrapper.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -90,6 +90,10 @@ tasks.withType(GroovyCompile) {
 }
 bootJar.enabled = false
 
+tasks.bootRun.doFirst {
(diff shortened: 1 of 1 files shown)


amcclain:
Running through a PR as we might want to look / test a bit further - perhaps we could remove/disable the task, or (to the contrary) perhaps the task is important in some other context that this change would break...

But wanted to give devs some help, as the failure pattern is pretty obscure if you run from the wrapper, which is itself easy to do.

lbwexler:
I like this change -- willing to try it!

PR #239, opened 2022-04-12
Made HOIST_ADMIN/listfiles also return file size and last modified date

jskupsik:


lbwexler (approved):
This looks fine, and certainly going to be compatible with earlier versions of hoist-react.

Note my minor tweak to the groovy syntax to be a bit more standard.  Happy to discuss the motivation there.

amcclain:
@jskupsik can you please add a CHANGELOG entry for this one so we can get it merged?  Would full under a "New Feature" heading.

PR #240, opened 2022-04-18
add missing setting to support grid exports up to 20MB

cnrudd:


src/main/groovy/io/xh/hoist/configuration/ApplicationConfig.groovy
```diff
@@ -91,6 +92,13 @@ class ApplicationConfig {
                 }
                 show_sql = false
             }
+
+            // Increase max form post size to 20mb to support large grid exports
+            server {
+                tomcat {
+                    maxHttpFormPostSize = 20971520
```
amcclain, line 99: Thanks for setting this up.  Given the recent discussions around _even bigger_ exports in our client projects, should we consider raising this further, e.g. to 50 or 100mb?

cnrudd: @amcclain NP.  After checking toolbox in production, I see that this setting is already set here:
https://github.com/xh/xh-tomcat/blob/jdk17/conf/server.xml#L73

but it only works the app is deployed.

So, this PR is not a  "must merge", but if we don't merge it, we might consider adding more comments telling devs where this setting can set to help with local development, 
or how to override it in either their app's application.groovy or tomcat/Dockerfile.

cnrudd: If we do decide to raise the limit to 50mb, we should also consider raising the timeout on exporAsync (js method) from 30 sec to 1.5mn.

cnrudd: I do wonder if this is the right place for adjusting an export size: https://github.com/xh/xh-tomcat/blob/jdk17/conf/server.xml#L73 when it can be done in the app's groovy code.  
Since looks like adjusting export limits requires 2 config changes, and if they can both be done in the same file, I think that is better than one in the groovy, one in docker/tomcat config.

amcclain: I was going to ask - how does this relate to the Tomcat setting? I would much rather have the app setting and not the Tomcat one, which is more obscure and needs to be kept manually up to date.  Are we confident that this does the same thing / could replace the Tomcat setting entirely?  

If so, we could remove from Tomcat and use that as an excuse to release a properly versioned new build of the base image there, while now setting (and I would vote for increasing) the limit here.

cnrudd: In an embarrassing revelation, it turns out I never confirmed that the custom export endpoint in the main app supports form posts > 20mb.   You can't easily test this in this client app, and I had just assumed that it had at one time been confirmed to work.

Petra just found a way to test this and confirmed that it does not work, even though the app has this setting in application.groovy:
```
    // Increase max post size for large grid exports
    server.tomcat.maxHttpPostSize = 52428800
```

So, it turns out the issue is:
setting `server.tomcat.maxHttpPostSize`  (or `server.tomcat.maxHttpFormPostSize` in Grails 5) does not override what is set in the Tomcat `server.xml`.

I vote for releasing a new build of the base image with `maxPostSize="20971520"` removed, so that we can test/confirm that this setting in application.groovy:
```
    // Increase max post size for large grid exports
    server.tomcat.maxHttpPostSize = 52428800
```
will work.


If we decide to increase the default size, we should also match the increase here:
https://github.com/xh/xh-nginx/blob/develop/xh.conf#L15

petradish: I wonder whether the default `maxPostSize` config for tomcat (2MB) will still not be overridden by our application config?

cnrudd: @petradish my guess is the default 2mb will be overrriden, because it is overriden when we run with
```
    // Increase max post size for large grid exports
    server.tomcat.maxHttpPostSize = 52428800
```
on our local dev machines.

petradish: I think that sounds right -- would the override for the tomcat config be set in the helm chart then?

cnrudd:
Closing.  The issue was resolved by increasing the maxPostSize value in [this project's](https://github.com/xh/xh-tomcat) server.xml files

PR #243, opened 2022-04-23
Excel export should only coerce explicit 'number' type values

jskupsik:


lbwexler:
Made some comments on the related hoist-react PR, with the assumption that this 'type' string be more clearly provided as null, or one of our `FieldType` enumeration.

It seems to me, that if we have type= `null` or type = `auto`, we could still go ahead and try to convert the numbers.    Not really sure, and we could go either way.  I realize we are frowning on  use of 'auto', but I think if it is used, we are going to help more people by converting numbers to numbers, than we will harm by converting the occasional string to a number.

grails-app/services/io/xh/hoist/export/GridExportImplService.groovy
```diff
@@ -186,16 +190,28 @@ class GridExportImplService extends BaseService {
...
-                // Collect cell value and cell format
-                def value, format
+                // Collect cell value, format, and type
+                def value, format, type
                 if (data instanceof Map) {
                     value = data?.value
-                    format = data?.format
+                    format = data?.format ?: metadata.format
+                    type = data?.type ?: metadata.type
                 } else {
```
petradish, line 199: I don't think this if/else block is needed, since `forExcel` on client will always be true here

grails-app/services/io/xh/hoist/export/GridExportImplService.groovy
```diff
@@ -186,16 +190,28 @@ class GridExportImplService extends BaseService {
...
+                    type = metadata?.type
                 }
+
                 value = value?.toString()
+                format = format?.toString()
+                type = type?.toString()
+
+                // Set a default excel format for DATE and LOCAL_DATE typed values
+                if (type == 'localDate' && (format == null || format == 'General')) {
+                    format = 'yyyy-mm-dd'
```
petradish, line 211: I think it would be nice if these strings were constants or part of some ExcelFormat enum, to mirror what we're doing on the client.

grails-app/services/io/xh/hoist/export/GridExportImplService.groovy
```diff
@@ -186,16 +190,28 @@ class GridExportImplService extends BaseService {
...
+                    type = data?.type ?: metadata.type
                 } else {
                     value = data
-                    format = metadata.format
+                    format = metadata?.format
+                    type = metadata?.type
                 }
+
                 value = value?.toString()
+                format = format?.toString()
```
petradish, line 206: Here and below, I think it would be good if format was never null (default to DEFAULT/'General')

grails-app/services/io/xh/hoist/export/GridExportImplService.groovy
```diff
@@ -213,22 +229,21 @@ class GridExportImplService extends BaseService {
...
-                            value = Date.parse('yyyy-MM-dd HH:mm:ss', value)
-                        } else if (metadata.type == 'int' || (!metadata.type && value.isInteger())) {
-                            value = value.toInteger()
-                        } else if (metadata.type == 'double' || (!metadata.type && value.isDouble())) {
+                        if (type == 'localDate' || format == 'yyyy-mm-dd') {
+                            value = LocalDate.parse(value)
+                        } else if (type == 'date' || format == 'yyyy-mm-dd h:mm AM/PM') {
+                            value = new SimpleDateFormat().parse(value)
+                        } else if (type == 'int' || type == 'number' || format?.contains('0')) {
                             value = value.toDouble()
```
petradish, line 239: I'm not certain if this is desired behavior for integers - I believe coercing to a double will set a decimal place after the number. Probably safer to leave the raw value as an integer.

grails-app/services/io/xh/hoist/export/GridExportImplService.groovy
```diff
@@ -213,22 +229,21 @@ class GridExportImplService extends BaseService {
...
-                            value = Date.parse('yyyy-MM-dd', value)
-                        } else if (metadata.type == 'datetime') {
-                            value = Date.parse('yyyy-MM-dd HH:mm:ss', value)
-                        } else if (metadata.type == 'int' || (!metadata.type && value.isInteger())) {
-                            value = value.toInteger()
-                        } else if (metadata.type == 'double' || (!metadata.type && value.isDouble())) {
+                        if (type == 'localDate' || format == 'yyyy-mm-dd') {
+                            value = LocalDate.parse(value)
+                        } else if (type == 'date' || format == 'yyyy-mm-dd h:mm AM/PM') {
+                            value = new SimpleDateFormat().parse(value)
```
petradish, line 237: I am not familiar enough with SimpleDateFormat to know if the default pattern is what we want - is there any reason to change this from the previous Date.parse('yyyy-MM-dd HH:mm:ss', value)?

petradish (changes requested):
Hey Jakub! I think this is a step in the right direction! Please see comments below and let me know if you have any questions!

grails-app/services/io/xh/hoist/export/GridExportImplService.groovy
jskupsik: I reverted back to the old behavior, since there is no need for data to be an object if there isn't a cell-level export definition. Type only varies by column so it never has to be passed by a cell. 

Data should should usually just be a value again.

grails-app/services/io/xh/hoist/export/GridExportImplService.groovy
```diff
@@ -8,6 +8,8 @@
 package io.xh.hoist.export
 
 import io.xh.hoist.BaseService
+import org.apache.poi.sl.usermodel.TextRun
```
petradish, line 11: Let's remove these extra imports here and line 30

src/main/groovy/io/xh/hoist/export/ExcelFormat.groovy
```diff
@@ -0,0 +1,61 @@
+package io.xh.hoist.export
```
cnrudd, line 1: This new file is awesome.  Thanks for adding this!

src/main/groovy/io/xh/hoist/export/FieldType.groovy
```diff
@@ -0,0 +1,30 @@
+package io.xh.hoist.export
```
cnrudd, line 1: Also great addition!

cnrudd (approved):

grails-app/services/io/xh/hoist/export/GridExportImplService.groovy
```diff
@@ -186,16 +195,22 @@ class GridExportImplService extends BaseService {
...
-                // Collect cell value and cell format
-                def value, format
+                // Collect cell value, format, and type
+                def value, format, type
                 if (data instanceof Map) {
-                    value = data?.value
-                    format = data?.format
+                    value = data.value
+                    format = data.format ?: metadata.format
+                    type = data.type ?: metadata.type
```
TomTirapani, line 203: As explained in the hoist-react PR, I think we can drop the handling of cell-level type and just use the metadata type

TomTirapani (approved):
Looks great! I pushed some small changes in accordance to my comments, and think this is ready to be merged :)

PR #245, opened 2022-05-20
Impersonation + Server Restart bug fix

jskupsik:
Hard to find but an easy fix. 

We were accidentally setting the session[APPARENT_USER_KEY] to the authUser object itself rather than its authUser.username string. This object was being converted into the desired username string so the bug was not apparent until server restart. The intended string values of session[APPARENT_USER_KEY] and session[AUTH_USER_KEY] are being properly persisted through a server restart, but the erroneous object value of session[APPARENT_USER_KEY] was not persisted during a server restart and was defaulting to null, resulting in the NullPointerException error.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## 14.0.0-SNAPSHOT - unreleased
+
+### 🐞 Bug Fixes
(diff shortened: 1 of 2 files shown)


lbwexler:
Wow!  Great find!  Very subtle, due to the toString() method on HoistUser

PR #253, opened 2022-07-22
`DateTimeUtils` tweak to support no-arg calls

amcclain:
+ App/server timezone conversion utils default to current day/date if called without arguments.
+ Added static getAppZoneId / getServerZoneId convenience getters

### DIFF
--- a/src/main/groovy/io/xh/hoist/util/DateTimeUtils.groovy
+++ b/src/main/groovy/io/xh/hoist/util/DateTimeUtils.groovy
@@ -11,6 +11,7 @@ import groovy.transform.CompileStatic
 
 import java.time.LocalDate
 import java.time.LocalTime
+import java.time.ZoneId
 
 import static java.time.format.DateTimeFormatter.BASIC_ISO_DATE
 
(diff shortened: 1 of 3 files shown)


src/main/groovy/io/xh/hoist/util/DateTimeUtils.groovy
```diff
@@ -43,32 +44,40 @@ class DateTimeUtils {
...
+        return serverTimeZone.toZoneId()
+    }
+
+    static LocalDate appDay(Date forDate = null) {
+        forDate ? forDate.toInstant().atZone(appZoneId).toLocalDate() : LocalDate.now(appZoneId)
     }
 
-    static LocalDate serverDay(Date dt) {
-        dt.toInstant().atZone(serverTimeZone.toZoneId()).toLocalDate()
+    static LocalDate serverDay(Date forDate = null) {
```
lbwexler, line 63: seems like this might be more self documenting and implementation simpler if it was forDate = new Date() -- but there are pro's and cons

PR #254, opened 2022-07-23
Optimize `preService.clientConfig`

amcclain:
+ With SQL logging enabled, I noticed we were issuing one `select` per non-local preference when a user visited the app.
+ Second-level query caching mitigated this, but only after a given user had visited at least once to populate the cache for their username.
+ Now only a single `select` is issued to fetch all user prefs at once.

### DIFF
--- a/grails-app/services/io/xh/hoist/pref/PrefService.groovy
+++ b/grails-app/services/io/xh/hoist/pref/PrefService.groovy
@@ -90,21 +90,18 @@ class PrefService extends BaseService {
 
     @Transactional
     void clearPreferences(String username = username) {
-        UserPreference.findAllByUsername(username).each {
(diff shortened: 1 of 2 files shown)


grails-app/services/io/xh/hoist/pref/PrefService.groovy
```diff
@@ -90,21 +90,18 @@ class PrefService extends BaseService {
 
     @Transactional
     void clearPreferences(String username = username) {
-        UserPreference.findAllByUsername(username).each {
-            UserPreference userPref = (UserPreference) it
-            userPref.delete()
-        }
+        getUserPrefs(username).each { it.delete() }
```
amcclain, line 93: Tested this and did not run into any issues.

PR #259, opened 2022-07-28
Standardize JSON serialization of date-like objects

lbwexler:


src/main/groovy/io/xh/hoist/json/JSONParser.java
```diff
@@ -25,10 +29,16 @@ public class JSONParser {
...
+
+        mapper = new ObjectMapper()
+                .registerModule(javaTimeModule)
+                .disable(READ_DATE_TIMESTAMPS_AS_NANOSECONDS);
 
         validateMapper = new ObjectMapper()
-                .enable(DeserializationFeature.FAIL_ON_TRAILING_TOKENS);
+                .registerModule(javaTimeModule)
+                .disable(READ_DATE_TIMESTAMPS_AS_NANOSECONDS)
+                .enable(FAIL_ON_TRAILING_TOKENS);
```
cnrudd, line 41: From what I read on the [docs/code ](https://github.com/FasterXML/jackson-databind/search?p=2&q=FAIL_ON_TRAILING_TOKENS)of this: FAIL_ON_TRAILING_TOKENS, I wonder if turning this on would prevent the framework in the future from reading in JSON as a stream?  Since it "forces reading of the whole input as single value?"   If we didn't need it before, why do we need it now?

cnrudd (commented):
I left one question about FAIL_ON_TRAILING_TOKENS.  For the rest I have discussed with Lee.  This will allow us to remove some code from a client App, which is great.

src/main/groovy/io/xh/hoist/json/JSONParser.java
amcclain: Looks like we did have it before?

amcclain (approved):
Looks like a great clean-up to me 👍

PR #260, opened 2022-07-29
Raise controller size limits to match xh-tomcat settings

cnrudd:
see: https://github.com/xh/xh-tomcat/blob/develop/conf/server.xml#L73

### DIFF
--- a/src/main/groovy/io/xh/hoist/configuration/ApplicationConfig.groovy
+++ b/src/main/groovy/io/xh/hoist/configuration/ApplicationConfig.groovy
@@ -70,10 +70,10 @@ class ApplicationConfig {
                 controllers {
                     defaultScope = 'singleton'
 
-                    // Increase limits to 20mb to support large grid exports, other file uploads.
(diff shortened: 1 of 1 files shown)


lbwexler:
Hi -- according to @Colin and @Petra, This change would align us on a single value that we will actually support by default for our upload size.

Question is, do we feel comfortable with having such a large value by default, or would we prefer to lower the value that we have in the server.xml file?

Or is what we have the best case scenario -- the difficult to modify server.xml setting is fairly high, but the easy to modify application.config setting is fairly low, but can be easily raised, e.g. :
```
grails.controllers.upload.maxFileSize = 52428800
grails.controllers.upload.maxRequestSize = 52428800
```

lbwexler:
There are comments related to this in an earlier PR -- attaching here for reference.
@amcclain seeming to favor the larger default values:
https://github.com/xh/hoist-core/pull/240

lbwexler:
we discussed this on slack a few weeks back-- easy enough to raise this if needed in client plugins/apps, but safer to leave this at the lower value.  closing.

PR #269, opened 2022-09-30
Monitor message will be prefixed instead of overriden on failure

### LINKED ISSUE #268: Status monitor should not overwrite custom message with its generated one
The status monitor API allows a check to set `MonitorResult.message` directly during execution, but `MonitorResultService` will overwrite that message in a variety of failure scenarios.

Sometimes the check might wish to use metric-driven evaluation (i.e. not set the result itself directly), but still has something useful to say even if the returned metrics eval to a failure.

To improve, the service could prepend its own generated message onto any message set by the check, instead of overwriting it completely. (I.e. the generated message is still useful / important, so show it in addition to any custom msg)

jskupsik:
Small change to the way the Monitor sets the results.message variable on failure. It will now prepend the error, warn, or fail message but still always render the user-provided message afterwards, rather than overriding it. 

Backwards and forwards compatible:
Old hoist-core + new hoist-react = appears identical to old behavior, the non-success message overrides any user-provided
New hoist-core + old hoist-react = the messages on the monitor card are displayed on the same line, rather than a newline

Closes #268

### DIFF
--- a/grails-app/services/io/xh/hoist/monitor/MonitorResultService.groovy
+++ b/grails-app/services/io/xh/hoist/monitor/MonitorResultService.groovy
@@ -109,7 +109,7 @@ class MonitorResultService extends BaseService {
 
         if (metric == null) {
             result.status = FAIL
-            result.message = "Monitor failed to compute metric \n\n$result.message"
+            result.prependMessage("Monitor failed to compute metric")
             return
         }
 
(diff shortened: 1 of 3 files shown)


grails-app/services/io/xh/hoist/monitor/MonitorResultService.groovy
```diff
@@ -108,7 +109,7 @@ class MonitorResultService extends BaseService {
 
         if (metric == null) {
```
lbwexler, line 110: why not pre-pending in this case?

lbwexler (commented):
Looks good -- let me know about the one question....

amcclain (approved):

PR #271, opened 2022-10-17
New role: HOIST_ADMIN_READER

cnrudd:


PeteDarinzo (approved):
Reviewed and discussed.

grails-app/controllers/io/xh/hoist/security/AccessInterceptor.groovy
```diff
@@ -38,32 +39,39 @@ class AccessInterceptor {
...
-        }
+        String actionNm = actionName ?: controllerClass.defaultAction
+        Method method = clazz.getMethod(actionNm)
 
-        Access access = clazz.getAnnotation(Access)
-        if (access) {
+        def access = method.getAnnotation(Access) ?:
+                method.getAnnotation(AccessAll) ?:
+                clazz.getAnnotation(Access) as Access ?:
+                clazz.getAnnotation(AccessAll) as AccessAll
```
amcclain, line 48: Curious why the cast is needed on the clazz-level calls but not the method ones.

grails-app/services/io/xh/hoist/user/BaseRoleService.groovy
```diff
@@ -56,9 +56,16 @@ abstract class BaseRoleService extends BaseService {
...
     Set<String> getRolesForUser(String username) {
-        return allRoleAssignments
+        Set<String> roles = allRoleAssignments
             .findAll{role, users -> users.contains(username)}
             .keySet()
+            .toSet()
+
+        if (roles.contains('HOIST_ADMIN')) {
+            roles.add('HOIST_ADMIN_READER')
+        }
```
amcclain, line 66: This method (`getRolesForUser`) is the primary one, as `HoistUser.getRoles()` delegates here.

There could be some edge-case bugs reported where an app has entirely overridden this method (which the doc comment makes clear it could do) or has other code that relies on `getAllRoleAssignments` or `getUsersForRole`, neither of which will have this behavior.

Seems like here we are weighing a tradeoff between building in the inheritance here, in a place where it will almost always have the right effect but could be inconsistent - vs requiring apps to update their role handling to ensure HOIST_ADMIN inherits HOIST_ADMIN_READER and handle that perm like all the others.  

I am not sure where I stand on it - not opposed to this pragmatic approach, but would want to at minimum document a decision to only handle here via a comment.

amcclain: Also - let's use the statics `HoistUser.HOIST_ADMIN_ROLE` + READER equiv

amcclain (commented):
Such a great pickup to get the per-method `@Access` annotations - in addition to the admin_reader feature that prompted this PR. 👍 

Please see questions below - none of which are blocking IMO.

grails-app/controllers/io/xh/hoist/security/AccessInterceptor.groovy
cnrudd: We noticed this too and tried to remove as Access.  We got a compilation error.

grails-app/services/io/xh/hoist/user/BaseRoleService.groovy
cnrudd: @amcclain re: "but would want to at minimum document a decision to only handle here via a comment."

Are you saying that if we keep the logic as is, an additional comment like: "If an application overrides this methos, it MUST recreate the logic that adds the HOIST_ADMIN_READER role to the HOIST_ADMIN role" should be added?

amcclain: Yes - I was looking for something to explicitly acknowledge in the code that there is room for inconsistency here and that we're applying this inheritance in one place but not others and in a way that would need to be replicated by an app that overrides this method.  I don't feel super-great about it and will wonder if the tradeoff is worth it.

amcclain: > @amcclain re: "Also - let's use the statics HoistUser.HOIST_ADMIN_ROLE + READER equiv" I had tried that. I was not able to import HoistUser.HOIST_ADMIN_ROLE here. Maybe because HoistUser is a trait?

If there is some unexpected / subtle problem with us having these static values hang off of the trait, I vote we move them somewhere else so they can be used as intended throughout the codebase.  I don't think it's essential that we have static values for these strings, but it seems like a good idea, and it *is* weird to have static values defined and then not be able to use them in the way we intended. 😄

lbwexler:
Merging this -- we will make additional changes to changelog, and documentation of BaseRoleService before release.

I don't actually think this needs to be considered a breaking change.

PR #273, opened 2022-10-31
Improve LogSupport signatures w/generics

amcclain:
+ Avoid need to cast return types out of `withInfo` and friends.
+ Changelog formatting for consistent line length, .editorconfig to set automatically.


lbwexler:
Great pick up

PR #279, opened 2023-03-12
Unpin Logback version

lbwexler:
We can take this now, but note that it will require our applications to rename their `logback.groovy` files.  There might need to be an additional rename back to `logback.groovy` in the future, which would be annoying.

Alternatively we could wait a tick, untils grails upgrades to the lastest version of logback, and it looks like then the file rename would not be needed.   Guess it depends on how anxious we are to unping logback.groovy.

See matching toolbox branch here

https://github.com/xh/toolbox/pull/644


CHANGELOG.md
```diff
@@ -124,6 +124,14 @@
 
 * grails `5.2.1 -> 5.3.2`
 
+### ⚙️ Breaking Changes
+* v16 restores the use of the standard grails version of logback (v1.2.11).  Hoist had been
+pinned on an earlier version of logback due to dropped support of groovy based configuration.
+This support is now restored via plugin. This change will require renaming the
+`conf/logback.groovy` file in your application to `conf/logback-config.groovy`.  It also may
```
amcclain, line 131: Note that this is no longer the case - can stick with logback.groovy as per updated dogbert docs.

build.gradle
```diff
@@ -76,6 +76,7 @@ dependencies {
     api 'org.jasypt:jasypt:1.9.3'
     api "commons-io:commons-io:2.8.0"
     api "org.owasp.encoder:encoder:1.2.3"
+    implementation 'io.github.virtualdogbert:logback-groovy-config:1.0'
```
amcclain, line 79: There are now different versions of this library depending on the particular minor version of logback we're looking to take.

amcclain (commented):
Apols for letting this linger for so long w/o comment.  Ryan just ran into this while updating a Hoist app - needed to add the magic `logback.version=1.2.7` to `gradle.properties`.  

Not a big hassle, but one more bit of boilerplate spread across apps, and the kind of thing that will likely get caught in various scans, etc. going forward. So am motivated to merge if we have a look at the latest from dogbert and determine it's all working as expected.

amcclain:
(Happy to try to make the updates to the library version and tweak this PR - figured we could decide if we have any interest in doing that before the big multi-instance merge, or not)

lbwexler:
I looked at this a month or two back, but was frustrated to see that grails was still on a newer, but *not* latest version of logback. That version would require apps to rename the `logback.groovy` with this workaround.  Unfortunately, we are unable to take the very latest version of logback as it is incompatible with grails.

Chatted with the author of the plugin (formerly known as virtualdogbert, now groovyduke) and he suggested we wait until grails upgrades logback to the latest.

CHANGELOG.md
lbwexler: Unfortunately, not that simple (see below)

PR #281, opened 2023-03-24
Safer TrackLog creation/querying

amcclain:
New `xhActivityTrackingConfig` w/constraints for safer TrackLog creation/querying

+ Partial mitigation of a major TrackService pitfall - querying and serializing large numbers of TrackLogs that themselves contain large nested `data` maps requires a significant allocation of memory and in the worst case can crash the server.
+ New config limits maximum size of tracked data and provides configurable (and lowered) constraints on admin client querying.
+ Intended for use with corresponding hoist-react v56 updates to respect the same config on the client.

### DIFF
--- a/grails-app/services/io/xh/hoist/track/TrackService.groovy
+++ b/grails-app/services/io/xh/hoist/track/TrackService.groovy
@@ -10,6 +10,7 @@ package io.xh.hoist.track
 import grails.events.EventPublisher
 import groovy.transform.CompileStatic
 import io.xh.hoist.BaseService
+import io.xh.hoist.config.ConfigService
 
 import static io.xh.hoist.browser.Utils.getBrowser
 import static io.xh.hoist.browser.Utils.getDevice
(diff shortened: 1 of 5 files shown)


grails-app/services/io/xh/hoist/track/TrackService.groovy
```diff
@@ -22,17 +23,21 @@ import static io.xh.hoist.util.Utils.getCurrentRequest
...
- * viewable via the Hoist Admin Console's Client Activity > Activity grid.
+ * viewable via the Hoist Admin Console's Activity > Tracking tab.
  *
  * The choice of which activities to track is up to application developers. Typical use-cases
  * involve logging queries and tracking if / how often a given feature is actually used.
  *
- * Persistence of track logs to the database can be disabled (e.g. in a local dev environment)
- * with an instance config value of `disableTrackLog: true`.
+ * The `xhActivityTrackingConfig` soft-config can be used to configure this service, including
+ * disabling it completely. Separately, the `disableTrackLog` *instance* config can be used to
```
lbwexler, line 32: given that we already have this, am wondering if we should skip the soft Config for disabling.  Or vice versa.  Seems like a lot of cognitive load here around disabling tracking.

amcclain: Please see upcoming client-side PR

PR #282, opened 2023-03-31
Added new `io.xh.hoist.data.filter` package

amcclain:
+ Basic implementations of data-filtering classes and utils - (partially) mirroring their TypeScript equivalents within Hoist React.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,11 +2,17 @@
 
 ## 17.0-SNAPSHOT - unreleased
 
+### 🎁 New Features
 
+* Added new `io.xh.hoist.data.filter` package, with basic implementations of data-filtering classes
+  and utils mirroring their TypeScript equivalents within Hoist React.
+    * Note these new classes are currently incubating, do not yet implement all features of their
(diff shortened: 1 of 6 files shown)


amcclain:
Before merging, would like to discuss how we might put some tests in to keep these honest and confirm some of the subtleties. 

This is such a clear example of utility code that could be tested in isolation, I feel like we really should add something to put a stake in the ground for hoist-core unit testing where we can do it.

TomTirapani:
Code and comments look great 👍 

Agree that this would be a great candidate for unit testing - is unit testing we've something we've looked into before, or have wider plans about? Looks like Grails has some built-in support for testing (https://testing.grails.org/latest/guide/index.html), but we could also look into lower-level pure groovy / java testing since these are in `/src/`

lbwexler:
are we still going ahead with this, 2 months later?  Sorry for the delay here

amcclain:
Looks like this will very likely be merged as part of #341 , which is great.  We can close out this PR and delete the original branch as soon as that happens.

PR #283, opened 2023-04-04
MemoryMonitoringService enhancements

lbwexler:
Enhance MemoryMonitor:

+ Add GC statistics
+ Add ability to take heap dump
+ Add pctMax as primary metric

### DIFF
--- a/grails-app/services/io/xh/hoist/monitor/MemoryMonitoringService.groovy
+++ b/grails-app/services/io/xh/hoist/monitor/MemoryMonitoringService.groovy
@@ -52,11 +52,14 @@ class MemoryMonitoringService extends BaseService {
      * Dump the heap to a file for analysis.
      */
     void dumpHeap(String filename) {
-        def heapDumpDir = config.heapDumpDir
-        if (heapDumpDir == null) {
+        String heapDumpDir = config.heapDumpDir
+        if (!heapDumpDir) {
             throw new RuntimeException(
                 "Unable to dump heap. Please specify value for 'xhMemoryMonitor.heapDumpDir'"
             )
+        }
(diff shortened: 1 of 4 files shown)


grails-app/services/io/xh/hoist/monitor/MemoryMonitoringService.groovy
```diff
@@ -42,13 +47,24 @@ class MemoryMonitoringService extends BaseService {
...
     }
 
+    /**
+     * Dump the heap to a file for analysis.
+     */
+    void dumpHeap() {
+        def filename = "heapdump_${currentTimeMillis()}.hprof"
+        withInfo('Dumping Heap') {
+            HotSpotDiagnosticMXBean mxBean = ManagementFactory.getPlatformMXBean(HotSpotDiagnosticMXBean.class)
+            mxBean.dumpHeap(filename, true)
```
haynesjm42, line 57: Wondering what impact this might have on the container filesystem - also based on the expected size of this file could we potentially send this up to the client to download in the browser? (and then delete the file from the container fs)

lbwexler:
CR: John H

grails-app/services/io/xh/hoist/monitor/MemoryMonitoringService.groovy
```diff
@@ -7,15 +7,20 @@
 
 package io.xh.hoist.monitor
 
+import com.sun.management.HotSpotDiagnosticMXBean
```
amcclain, line 10: Is this class available in all vendor-specific flavors of the JDK?

lbwexler:
it's a great question -- some form of it should be, but will double check,
and make sure that this is the appropriate interface to refer to it with.

PR #289, opened 2023-05-22
New DateTimeUtils.ensureServerTimeZoneIs() util method

### LINKED ISSUE #215: Verification for Server Timezone
Would like to provide a config driven capability to verify that the JVM timezone is what we expect it to be.  Optional, and can specify either that a warning should be triggered, or the app should fail startup.

Important for clients that really require a certain timezone due to e.g. timezone specific data storage.

Don't want to try to programmatically set the timezone, because we really want the right timezone from the get go of JVM startup, logging, etc.

amcclain:
+ Callable from e.g. app Bootstrap routines to enforce a particular time zone, if required by the app (e.g. due to interactions with a database running in a particular zone).
+ Log default zone in Hoist banner message on startup.
+ Fixes #215

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,9 +4,13 @@
 
 ### 🎁 New Features
 
-* Add support for the `caseSensitive` flag in log filtering endpoint.
+* Added support for the `caseSensitive` flag in log filtering endpoint.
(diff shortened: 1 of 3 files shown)


amcclain:
Thanks for merging!

PR #292, opened 2023-05-31
Banner presets

leogretz2:


grails-app/services/io/xh/hoist/alertbanner/AlertBannerService.groovy
```diff
@@ -70,6 +75,24 @@ class AlertBannerService extends BaseService {
...
 
+    List getAlertPresets() {
+        def svc = jsonBlobService,
+            presetsBlob = svc.list(blobType, blobOwner).find { it.name == presetsBlobName }
+
+        presetsBlob ? JSONParser.parseArray(presetsBlob.value) : []
+    }
+
+    void setAlertPresets(List presets) {
+        def svc = jsonBlobService,
```
lbwexler, line 86: I guess its ok to use the same blob 'type' for the banner and the presets.  I think our original vision was that all blobs of a given type would have the same structure.

grails-app/services/io/xh/hoist/alertbanner/AlertBannerService.groovy
```diff
@@ -33,9 +33,14 @@ class AlertBannerService extends BaseService {
     private final static String blobType = 'xhAlertBanner';
     private final static String blobOwner = 'xhAlertBannerService';
 
+    private final static String presetsBlobName = 'xhAlertBannerPresets';
+
     private final emptyAlert = [active: false]
     private Map cachedBanner = emptyAlert
 
+    private final emptyPresets = [active: false]
```
lbwexler, line 41: are these two properties on 41 and 42 obsolete/cruft?  Need to review all code after development to find stuff like this....

PR #293, opened 2023-06-06
New json parsing methods on BaseController

lbwexler:
+ Improved documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
 * Added new routine in `Bootstrap` to auto-create a new `xhExpectedServerTimeZone` app config.
   This config is now read at startup to validate that the server is running in the expected zone,
   and will throw a fatal exception if it is invalid or does not match the zone reported by Java.
+* Added new methods `parseJsonObject` and `parseJsonArray` to `BaseController`.
 
 ### 🐞 Bugfixes
 
--- a/grails-app/controllers/io/xh/hoist/BaseController.groovy
+++ b/grails-app/controllers/io/xh/hoist/BaseController.groovy
@@ -12,6 +12,7 @@ import grails.async.web.WebPromises
 import groovy.transform.CompileStatic
 import groovy.util.logging.Slf4j
 import io.xh.hoist.exception.ExceptionRenderer
+import io.xh.hoist.json.JSONParser
 import io.xh.hoist.json.JSONSerializer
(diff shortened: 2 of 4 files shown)


grails-app/controllers/io/xh/hoist/BaseController.groovy
```diff
@@ -25,11 +26,39 @@ abstract class BaseController implements IdentitySupport, LogSupport {
...
+        JSONParser.parseArray(request.inputStream)
+    }
+
+    /**
+     * Parse JSON submitted in the body of the request.
+     *
+     * Favor this method over the direct use of grails' request.getJSON() in order
+     * to utilize the customizable jackson-based parsing provided by Hoist.
+     */
+    protected Map parseJSONObject() {
```
cnrudd, line 58: if we went one step further on these and added request.InputStream as the default:
```protected Map parseJSONObject(obj = request.inputStream) {
        JSONParser.parseObject(obj)
    }```
then things like this on line 35 in RestController.groovy
JSONParser.parseObject(params.query)
could be:
parseJSONObject(params.query)

**Review by cnrudd — COMMENTED** [review]
Just one suggestion.

File: grails-app/controllers/io/xh/hoist/BaseController.groovy (hunk 1 above)
Comment (lbwexler) [reply grails-app/controllers/io/xh/hoist/BaseController.groovy:58]: could not do easily as the inputs are of two different types.  Really trying to provide just the replacement for request.getJSON() which is built in to grails.

File: CHANGELOG.md [hunk 2]
```diff
@@ -20,6 +20,7 @@
 * Added new routine in `Bootstrap` to auto-create a new `xhExpectedServerTimeZone` app config.
   This config is now read at startup to validate that the server is running in the expected zone,
   and will throw a fatal exception if it is invalid or does not match the zone reported by Java.
+* Added new methods `parseJsonObject` and `parseJsonArray` to `BaseController`.
```
amcclain, line 23: Capitalization mismatch here

amcclain:
Just saw that this change went in - thanks for taking it on. 

Trying to come up with a pair of method names that better communicate that these are parsing the request. Otherwise they seem to be a bit confusing/mysterious in the caller's context without any arguments, don't you think? What about `parseRequestJSON()` and `parseRequestJSONAsArray()`.

I don't personally think those are too long, but if we need to make them shorter we could sub `get` for `parse`, could drop the `As` in the second method name.

PR #295, opened 2023-06-14
Improve Impersonation

lbwexler:
+ Support for new 'HOIST_IMPERSONATOR' role
+ Allow customization of impersonation targets.

See matching PR in Hoist React

### DIFF
--- a/grails-app/services/io/xh/hoist/user/BaseRoleService.groovy
+++ b/grails-app/services/io/xh/hoist/user/BaseRoleService.groovy
@@ -65,6 +65,7 @@ abstract class BaseRoleService extends BaseService {
 
         if (roles.contains('HOIST_ADMIN')) {
             roles.add('HOIST_ADMIN_READER')
+            roles.add('HOIST_IMPERSONATOR')
         }
 
         return roles
--- a/grails-app/services/io/xh/hoist/user/BaseUserService.groovy
+++ b/grails-app/services/io/xh/hoist/user/BaseUserService.groovy
@@ -21,6 +21,20 @@ import io.xh.hoist.BaseService
  */
 @CompileStatic
 abstract class BaseUserService extends BaseService {
+    /** Return a user, or null if user not found. */
     abstract HoistUser find(String username)
+
+    /**
+     * Return all users,
+     * @param activeOnly - true to return "active" users only.
+     */
     abstract List<HoistUser> list(boolean activeOnly)
+
+    /**
+     * Return the users that a given user can impersonate.
+     * @param authUser
+     */
+    List<HoistUser> impersonationTargetsForUser(HoistUser authUser) {
+        authUser.canImpersonate ? list(true) : []
+    }
 }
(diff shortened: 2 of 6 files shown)


amcclain:
We should update the note installed on the `xhEnableImpersonation` config in Bootstrap - it still references Hoist Admins.  Won't update existing instances, but at least will be accurate going forward.

grails-app/services/io/xh/hoist/user/BaseRoleService.groovy
```diff
@@ -65,6 +65,7 @@ abstract class BaseRoleService extends BaseService {
 
         if (roles.contains('HOIST_ADMIN')) {
             roles.add('HOIST_ADMIN_READER')
+            roles.add('HOIST_IMPERSONATOR')
```
amcclain, line 68: A bit on the fence here, but overall I am leaning in this direction, to bake the inheritance in here.

(I see that we clearly state the method can be overridden, although I expect many apps will not have done so. It's not clear that all HOIST_ADMINs should always have impersonation ability, but for apps that are sourcing their roles from some external system, adding this in will minimize change at release.

grails-app/services/io/xh/hoist/user/BaseUserService.groovy
```diff
@@ -21,6 +21,20 @@ import io.xh.hoist.BaseService
...
+     * Return all users,
+     * @param activeOnly - true to return "active" users only.
+     */
     abstract List<HoistUser> list(boolean activeOnly)
+
+    /**
+     * Return the users that a given user can impersonate.
+     * @param authUser
+     */
+    List<HoistUser> impersonationTargetsForUser(HoistUser authUser) {
```
amcclain, line 37: Do we want any explicit mention of app-level overriding, to indicate that it's supported and the reason why we have this method here?  Or is that assumed to be the case?

grails-app/services/io/xh/hoist/user/IdentityService.groovy
```diff
@@ -74,27 +74,33 @@ class IdentityService extends BaseService {
...
+        def authUser = findAuthUser(request)
+        if (!authUser?.canImpersonate) {
+            throw new RuntimeException('User not found or user not authorized for impersonation.')
         }
-        if (!targetUser?.active) {
-            throw new RuntimeException("Cannot impersonate '$username' - no active user found")
+
+        def targetUser = userService.impersonationTargetsForUser(authUser).find {it.username == username}
+        if (!targetUser) {
+            throw new RuntimeException("'$authUser.username' does not have permissions to impersonate '$username'")
```
amcclain, line 103: Could just be `"$authUser does not..."` - toString prints username

grails-app/services/io/xh/hoist/user/BaseRoleService.groovy
lbwexler: yes, think this is very influenced by where we are, and minimizing change at release.   Would land differently if we had a clean slate.

grails-app/services/io/xh/hoist/user/BaseUserService.groovy
lbwexler: I think its pretty clear that this class exists for overriding.

grails-app/services/io/xh/hoist/user/IdentityService.groovy
lbwexler: nice, will do

PR #299, opened 2023-06-30
Version checking and forcing refresh

TomTirapani:
+ Deprecate config `xhAppVersionCheckEnabled` in favor of object based `xhAppVersionCheck`.
+ Add support for `shouldRequireRefresh` flag

See corresponding hoist-react PR: https://github.com/xh/hoist-react/pull/3407

### DIFF
--- a/grails-app/controllers/io/xh/hoist/impl/XhController.groovy
+++ b/grails-app/controllers/io/xh/hoist/impl/XhController.groovy
@@ -219,11 +219,11 @@ class XhController extends BaseController {
     }
 
     def version() {
-        def shouldUpdate = configService.getBool('xhAppVersionCheckEnabled')
-        renderJSON (
+        def options = configService.getMap('xhAppVersionCheck', [:])
+        renderJSON(
+            *: options,
             appVersion: Utils.appVersion,
-            appBuild: Utils.appBuild,
-            shouldUpdate: shouldUpdate
+            appBuild: Utils.appBuild
         )
     }
 
--- a/grails-app/init/io/xh/hoist/BootStrap.groovy
+++ b/grails-app/init/io/xh/hoist/BootStrap.groovy
@@ -59,8 +59,6 @@ class BootStrap {
     }
 
     private void ensureRequiredConfigsCreated() {
-        configService.configDeprecated('xhAppVersionCheckEnabled', 'v16.4.0', "Use 'xhAppVersionCheck' instead.")
-
         configService.ensureRequiredConfigsCreated([
             xhActivityTrackingConfig: [
                 valueType: 'json',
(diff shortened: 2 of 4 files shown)


grails-app/controllers/io/xh/hoist/impl/XhController.groovy
```diff
@@ -219,11 +219,11 @@ class XhController extends BaseController {
     }
 
     def version() {
-        def shouldUpdate = configService.getBool('xhAppVersionCheckEnabled')
-        renderJSON (
+        def options = configService.getMap('xhAppVersionCheck', [:])
```
lbwexler, line 222: shipping here to enable on-the-fly updates -- makes sense

grails-app/init/io/xh/hoist/BootStrap.groovy
```diff
@@ -92,11 +94,17 @@ class BootStrap {
                 groupName: 'xh.io',
                 note: 'Official TimeZone for this application - e.g. the zone of the head office. Used to format/parse business related dates that need to be considered and displayed consistently at all locations. Set to a valid Java TimeZone ID.'
             ],
-            xhAppVersionCheckEnabled: [
-                valueType: 'bool',
-                defaultValue: true,
+            xhAppVersionCheck: [
+                valueType: 'json',
+                defaultValue: [
+                    shouldUpdate: configService.getBool('xhAppVersionCheckEnabled'),
```
lbwexler, line 100: think we need a default -- brand new apps won't have the deprecated config

grails-app/init/io/xh/hoist/BootStrap.groovy
```diff
@@ -92,11 +94,17 @@ class BootStrap {
...
+                    shouldRequireRefresh: false
+                ],
                 groupName: 'xh.io',
-                note: 'True to show an update prompt banner when the server reports to the client that a new version is available. Can be set to false to temporarily avoid an upgrade prompt (e.g. while validating a deploy). Use xhAppVersionCheckSecs to completely disable this feature.'
+                note: 'Controls application behaviour when the server reports to the client that a new version is available. Supports the following options:\n\n' +
+                    '+ `shouldUpdate`: True to show an update prompt banner. Can be set to false to temporarily avoid an upgrade prompt (e.g. while validating a deploy).\n' +
+                    '+ `shouldRequireRefresh`: True to force clients to refresh immediately. Supersedes `shouldUpdate`. To be used when an updated server is known to be incompatible with a previously deployed client.\n\n' +
+                    'Use xhAppVersionCheckSecs to completely disable this feature.'
             ],
             xhAppVersionCheckSecs: [
```
lbwexler, line 109: should we deprecate this one while we are at it, and move it in to the bundle in the name of consolidation?

grails-app/services/io/xh/hoist/config/ConfigService.groovy
```diff
@@ -156,6 +156,19 @@ class ConfigService extends BaseService implements EventPublisher {
         logDebug("Validated presense of ${reqConfigs.size()} required configs", "created ${created}")
     }
 
+    /**
```
lbwexler, line 159: great idea -- should we just gather all of these in config.init()?   

Also, the message might be slightly different -- its not that the config is being removed.

grails-app/init/io/xh/hoist/BootStrap.groovy
TomTirapani: Good point - will default to true, to match the old config

grails-app/init/io/xh/hoist/BootStrap.groovy
TomTirapani: Could do, although the difference is `xhAppVersionCheckSecs` is available on the client, whereas `xhAppVersionCheck` is server-side only

grails-app/services/io/xh/hoist/config/ConfigService.groovy
TomTirapani: Spoke with Lee - decided to zap this, as potentially misleading for people who update hoist-core without hoist-react

grails-app/init/io/xh/hoist/BootStrap.groovy
```diff
@@ -93,18 +93,19 @@ class BootStrap {
...
-            xhAppVersionCheckSecs: [
-                valueType: 'int',
-                defaultValue: 30,
+            xhAppVersionCheck: [
+                valueType: 'json',
+                defaultValue: [
+                    interval: configService.getInt('xhAppVersionCheckSecs', 30),
+                    shouldUpdate: configService.getBool('xhAppVersionCheckEnabled', true),
+                    shouldRequireRefresh: false
+                ],
```
amcclain, line 102: I think it's a great goal to land on a consolidated config for this feature. The server-side configs are not sensitive, so mostly a matter of potential confusion where some keys are read on server and relayed to client, with others read by client directly. But having a single overall config to control the service seems like enough of a win to make it worth it.

What if we went with something like:
```
[
  // Read on client - will completely disable polling on client if false (i.e. really disable the feature, client would need to reload if changed).
  enabled: true,
  // Read on client - used to setup timer, but only if generally enabled.
  intervalSecs: 30,
  // Read on server, relayed to client with each check response.
  promptForReload: true,
  // Read on server, relayed to client with each check response.
  forceReload: false
]
```

Maybe we could introduce a level of nesting to pull out the two more dynamic configs, something like;

```
[
  enabled: true,
  intervalSecs: 30,
  dynamic: [
    promptForReload: true,
    forceReload: false
  ]
]
```

Where `dynamic` is some key name that someone can come up with that best expresses this idea.

lbwexler:
tom and I agreed on the single config and moving the interval in.

I think if we were going to so such aggressive renaming restructure (I am
sympathetic), I might prefer just a mode like

mode: 'forceReload/promptReload/silent'

Re: "silent" --  it seems like we should encourage that this timer to never
be turned off (or at least not advertise it with an explicit flag) as it
usefully updates the client environment service as to the new server
version, and having that always accurate could be useful for apps and hoist
alike

amcclain:
Yes I think the mode makes a lot of sense to eliminate possibility of contradictory options (prompt:false/force:true)

I was thinking we might run into a case where a client didn't want the polling (e.g. to avoid it acting as a session keepalive or something like that) - alert banners are the other example of constant background polling Hoist will do, and that does have such a flag.  But we could definitely cross that bridge if/when we ever come to it.

lbwexler:
if we ever came to that bridge, we could use -1 to get across it! 😀

PR #306, opened 2023-08-07
Admin log viewer log root path

leogretz2:


grails-app/controllers/io/xh/hoist/admin/LogViewerAdminController.groovy
```diff
@@ -18,15 +18,16 @@ class LogViewerAdminController extends BaseController {
     def logArchiveService,
         logReaderService
 
-    def listFiles() {
+    def listFileInfo() {
```
lbwexler, line 21: see my comments on hoist-react PR about leaving name the same.

PR #309, opened 2023-08-16
Hoist role support

lbwexler:
Server-side changes to support https://github.com/xh/hoist-react/pull/3447


grails-app/init/io/xh/hoist/BootStrap.groovy
```diff
@@ -309,6 +323,36 @@ class BootStrap {
...
     }
 
+    private void ensureRequiredRolesCreated() {
+        if (!configService.getMap('xhRoleModuleConfig').enabled || Role.count()) return
+        roleAdminService.ensureRequiredRolesCreated([
+            [
+                name: 'HOIST_ADMIN',
+                category: 'Hoist',
+                notes: 'Hoist Admins have full access to all Hoist Admin tools and functionality.',
+                users: ['ADMIN_USER']
```
lbwexler, line 333: Is this 'ADMIN_USER' something new we were contemplating?  Its not really a thing right now. :-)

ghsolomon: Intended to be a placeholder- idea being that a developer setting up a new app would get an initial admin `RoleMember` created for them, which they would replace with their username in order to gain initial access to the app.

PR #312, opened 2023-09-12
Added `ConfigService.updateConfig()`

jskupsik:


grails-app/services/io/xh/hoist/config/ConfigService.groovy
```diff
@@ -156,6 +155,48 @@ class ConfigService extends BaseService implements EventPublisher {
...
+     *      - updateConfig([name: 'xhExampleConfig', note: 'This is an example config.', value: 'Example value'])
+     *
+     * @param name OR params.name
+     * @param value OR params.value
+     * @param params.note
+     * @param params.lastUpdatedBy - defaults to authUsername or 'hoist-config-service'
+     * @returns - the updated AppConfig
+     */
+    @Transactional
+    AppConfig updateConfig(Map params, String name = params.name as String, Object value = params.value) {
```
amcclain, line 174: As discussed, this method signature seems very flexible, but also rather unusual.  I also wonder if the doc comment is accurately matching it - hard to say how it would be documented.  I'm always interested in using new / useful syntax, but the extra cognitive effort spent interpreting this flexible signature makes it a bit difficult to justify, especially as there are only a small number of parameters.  Would be great to discuss with @lbwexler .

grails-app/services/io/xh/hoist/config/ConfigService.groovy
```diff
@@ -156,6 +155,48 @@ class ConfigService extends BaseService implements EventPublisher {
...
+            note         : params.note as String,
+            lastUpdatedBy: params.lastUpdatedBy as String ?: authUsername ?: 'hoist-config-service'
+        ].each { k, v ->
+            // Update the provided fields - leave other fields alone
+            if (v != null) {
+                currConfig[k] = v
+            }
+        }
+
+        currConfig.save(flush: true)
```
amcclain, line 192: I'd suggest we skip `flush:true` here - in general the idea is to allow Hibernate to flush the session on its own accord, not sure we need to force it here.  Note we don't do so in createRequired... above.

grails-app/services/io/xh/hoist/config/ConfigService.groovy
```diff
@@ -156,6 +155,48 @@ class ConfigService extends BaseService implements EventPublisher {
...
+    @Transactional
+    AppConfig updateConfig(Map params, String name = params.name as String, Object value = params.value) {
+        def currConfig = AppConfig.findByName(name, [cache: true])
+
+        if (currConfig == null) {
+            throw new RuntimeException("No config found with name: [$name]")
+        }
+
+        [
+            value        : value,
```
amcclain, line 182: Should we add an explicit `toString()` here? Groovy must end up doing that implicitly when set into the `String value` property of AppConfig, eh?

But also what about json configs, where `value` would most naturally be provided as a map. We need to serialize it - as we do in ensureRequired... above.

amcclain (commented):
Pls see comments below - will also add @lbwexler as a reviewer to have a look so we can get this merged ASAP.

Lee - we have a use case for a client project where we need to support updating a small bit of global state related to a particular feature. AppConfig seemed like an appropriate option, but we need to add a managed / business-facing UI to set these options and therefore wanted a proper Hoist API to update the config itself on the back end.  

Happy to review the particulars in more detail, of course.

grails-app/services/io/xh/hoist/config/ConfigService.groovy
lbwexler: I am confused by this signature, but also think this is a case where simpler is better.  How often is this API really used?  We want to use these flexible API calls only for very heavily used APIs,  where the different variants are really going to save a lot of extra client code and so justify the complexity cost.

- Let's just go with something that aligns with the existing get methods, e.g. `setValue(name, value)`:   aligning your get and set methods is always a good goal.  

- Applications will rarely/never be changing the note, and we can typically intuit the user from identity service if it exists.  If we decide to support those, we could add an optional map at the end, but I would argue that we might wait on that

grails-app/services/io/xh/hoist/config/ConfigService.groovy
lbwexler: I think out standard for the one off GORM writes  that we may want to immediately read is  actually to use the `flush: true`.  May no longer be neccessary but it certainly is the standard, so would stick with it for consistency.   CrceateRequired is more part of a bulk process on startup, so arguably a different use-case.

PR #313, opened 2023-09-18
Simplify WebSocket push in (common) single channel case

lbwexler:
Tom and I were surprised to find this extra layer of async/thread usage, given how common this operation is


amcclain (approved):
Looks good to me - just reading the code here and don't have too much context around websockets in my mind at the moment, but certainly makes sense.

I checked and I see that we wrap the internals of `sendMessage()` in a try/catch, and our use of `ConcurrentWebSocketSessionDecorator` within `HoistWebSocketChannel` is setup with a configurable send timeout. 👍

TomTirapani:
Looks great, exactly as discussed 👍

PR #316, opened 2023-11-01
Improve Exception Handling in ProxyService

lbwexler:
In particular, don't want late exceptions in streaming to cause hard-to-debug exceptions from trying to change already committed response.

Phase 1 for improvements to ProxyService.  Phase 2 will be better default proxying of headers


lbwexler:
Greg -- hoping you can take a look as I know you are using this at a client and recently had some issues.

PR #323, opened 2024-01-18
Misc tweaks to DefaultRoleAdmin API

lbwexler:


src/main/groovy/io/xh/hoist/role/provided/DefaultRoleService.groovy
```diff
@@ -82,8 +78,8 @@ class DefaultRoleService extends BaseRoleService {
         defaultRoleAdminService
 
     private Timer timer
-    protected Map<String, Set<String>> _allRoleAssignments
-    protected ConcurrentMap<String, Set<String>> _roleAssignmentsByUser
+    protected Map<String, Set<String>> _allRoleAssignments = emptyMap()
```
ghsolomon, line 81: Super nit - since we're using `EMPTY_SET` in `getUsersForRole`, maybe would want to use `EMPTY_MAP` here?

src/main/groovy/io/xh/hoist/role/provided/DefaultRoleService.groovy
```diff
@@ -92,116 +88,95 @@ class DefaultRoleService extends BaseRoleService {
...
             Set<String> userRoles = new HashSet()
             allRoleAssignments.each { role, users ->
                 if (users.contains(username)) userRoles << role
             }
+            ret = _roleAssignmentsByUser[username] = unmodifiableSet(userRoles)
+        }
 
-            _roleAssignmentsByUser[username] = Collections.unmodifiableSet(userRoles)
+        if (getInstanceConfig('bootstrapAdminUser') == username && isLocalDevelopment && !isProduction) {
+            ret += ['HOIST_ADMIN', 'HOIST_ADMIN_READER', 'HOIST_ROLE_MANAGER']
```
ghsolomon, line 116: Might be missing something here, but isn't `ret` assigned an `unmodifiableSet` on line 112? Wondering how it can be mutated here

src/main/groovy/io/xh/hoist/role/provided/DefaultRoleService.groovy
lbwexler: will make these the same -- good point, I am very inconsistent here

PR #328, opened 2024-01-25
Fix `DefaultRoleService` unintended case-sensitive handling of usernames

ghsolomon:


src/main/groovy/io/xh/hoist/role/provided/DefaultRoleService.groovy
```diff
@@ -32,8 +32,8 @@ import static io.xh.hoist.util.InstanceConfigUtils.getInstanceConfig
  * Applications using this default implementation must either:
  *
  *  1) Define a `RoleService` class that extends this class. Allows for overriding its protected
- *     methods to customize behavior, including the option to implement `getUsersForDirectoryGroups`
- *     to resolve external directory group memberships (see below).
+ *     methods to customize behavior, including the option to implement
+ *     `doLoadUsersForDirectoryGroups` to resolve external directory group memberships (see below).
```
amcclain, line 36: Thanks for catching!

amcclain (commented):
Wondering if we want to take this further and have our `RoleMember` class normalize its `name` field to lowercase when `type == USER`.  We could do that using the GORM lifecycle hooks. It would get the data at rest into a more normal form and help ensure our GORM-level uniqueness constraint works as expected.

Likewise, we could leverage our `loadUsersForDirectoryGroups` wrapper around the app-supplied implementation method to lowercase all of the usernames emitted there, getting them into the lowercase format sooner so they show up that way (for better, I think) in the Admin UI.

Regardless, I think it's still the right choice to force to lowercase in the places you've added, to handle the small numbers of existing records that might already exist if nothing else.

lbwexler:
@amcclain -- was thinking about the same, but wondering whether clients that wanted to so see their usernames in the admin screen according to their own corporate conventions might prefer we not force the lowercasing at that stage.

Agree this is a good change in any case, even if we want to take it further at some point, and I will merge in prep for release

amcclain:
Got it, thanks.  I was a bit wary that some of the "effective" methods baked into `Role` could get thrown off by dupes with different casing, or something like that.  Not sure if that's a real issue

PR #329, opened 2024-01-26
Normalize role member usernames to lowercase and generally tighten up…

ghsolomon:
… case-insensitive handling


lbwexler:
Merging now to facilitate ongoing work on snapshot with client.  Review in develop pending

PR #333, opened 2024-02-07
Allow `InstanceConfig`s to be sourced from env vars + override `AppConfig`s

amcclain:


amcclain (commented):
Looking good to me, thank you @ghsolomon !

I pushed some tweaks to changelog/comment, as well as a more material change to DRY up the sensitive param detection into a new util method, which now also looks for `pwd` and `secret` and has an appropriately weaselly qualifying comment.

PR #339, opened 2024-03-20
Improve Directory Group lookup error handling

ghsolomon:


src/main/groovy/io/xh/hoist/role/provided/DefaultRoleService.groovy
```diff
@@ -269,18 +272,10 @@ class DefaultRoleService extends BaseRoleService {
...
-       if (!directoryGroups) return emptyMap()
-        // Wrapped here to avoid failing implementations from ever bringing down app.
-        try {
-            return doLoadUsersForDirectoryGroups(directoryGroups)
-        } catch (Throwable e) {
-            def errMsg = 'Error resolving Directory Groups'
-            logError(errMsg, e)
-            return directoryGroups.collectEntries {[it, errMsg] }
-        }
+        !directoryGroups ? emptyMap() : doLoadUsersForDirectoryGroups(directoryGroups)
```
lbwexler, line 275: perhaps inline the "do" method -- one less method to trace through

lbwexler (approved):

grails-app/services/io/xh/hoist/role/provided/DefaultRoleAdminService.groovy
```diff
@@ -41,9 +41,15 @@ class DefaultRoleAdminService extends BaseService {
         if (defaultRoleService.directoryGroupsSupported) {
             Set<String> groups = roles.collectMany(new HashSet()) { it.directoryGroups }
             if (groups) {
-                Map<String, Object> groupsLookup = defaultRoleService.loadUsersForDirectoryGroups(groups)
-                usersForGroups = groupsLookup.findAll { it.value instanceof Set }
-                errorsForGroups = groupsLookup.findAll { !(it.value instanceof Set) }
+                try {
+                    Map<String, Object> groupsLookup = defaultRoleService.loadUsersForDirectoryGroups(groups)
+                    usersForGroups = groupsLookup.findAll { it.value instanceof Set }
+                    errorsForGroups = groupsLookup.findAll { !(it.value instanceof Set) }
```
amcclain, line 47: To confirm - we might still get `errorsForGroups` populated as expected with partial lookup failures, but now only in the case of 'No AD group found' - is that correct?  

Checking that a search returnign no result does not throw- that would be reasonable but didn't want to assume.

(Would suggest changing to "Directory Group not found" FWIW)

src/main/groovy/io/xh/hoist/role/provided/DefaultRoleService.groovy
```diff
@@ -290,16 +285,22 @@ class DefaultRoleService extends BaseRoleService {
...
+                loadUsersForDirectoryGroups(groups).each { k, v ->
+                    if (v instanceof Set) {
+                        usersForDirectoryGroups[k] = v
+                    } else {
+                        logError("Error resolving users for directory group", k, v)
+                    }
                 }
+                _usersForDirectoryGroups = usersForDirectoryGroups
+            } catch (Throwable e) {
+                logError("Error resolving users for directory groups", e)
```
amcclain, line 303: Worth a comment to clarify that leaving the previous `_usersForDirectoryGroups` in place (and then re-processing it below) is a deliberate choice?

lbwexler:
Trying to think big picture here and wanted to get this out here.  In particular, don't want to trade one set of problems for another.

What motivated this change was the concern that we are making a large number of individual LDAP queries to get groups right now.  If any one of them fails (e.g. timeout, or just flaky) we can basically "blip out" user access for all of the users in the group that failed.  And we may be just doing our 10 minute refresh, replacing a perfectly good result with a flaky result -- and that seemed wrong.   So this should fix that.

New concern --  now a single flaky call keeps us from ever getting needed roles in the app. Or worse yet, someone enters something as a new or modified group tag that somehow causes all role groups to fail to load.

Maybe its overkill, but we could pass a flag to indicate whether we wanted "strict mode" or whether we are happy to have whatever results we can get?  The latter could be useful when we start-up, or for whatever reason have no successful load of groups.

grails-app/services/io/xh/hoist/role/provided/DefaultRoleAdminService.groovy
ghsolomon: Correct - will double-check this now to make sure. And sounds good re: change to error text

ghsolomon: Confirmed! Built a snapshot from this branch and tested in an EMC app to confirm.

lbwexler:
Created a new branch here to incorporate my thoughts above.  Greg reveiwed and testing out now

PR #341, opened 2024-03-21
Refactor TrackLogAdminController to support FieldFilter objects for server-side queries

jacob-xhio:
This PR addresses https://github.com/xh/hoist-react/issues/3572

### DIFF
--- a/grails-app/controllers/io/xh/hoist/admin/TrackLogAdminController.groovy
+++ b/grails-app/controllers/io/xh/hoist/admin/TrackLogAdminController.groovy
@@ -9,6 +9,8 @@ package io.xh.hoist.admin
 
 import grails.gorm.transactions.ReadOnly
 import io.xh.hoist.BaseController
+import io.xh.hoist.data.filter.Filter
+import io.xh.hoist.data.filter.Utils
 import io.xh.hoist.json.JSONParser
 import io.xh.hoist.security.Access
 import io.xh.hoist.track.TrackLog
@@ -40,7 +42,7 @@ class TrackLogAdminController extends BaseController {
 
         def filters = null
         if (params.filters) {
-            filters = JSONParser.parseObject(params.filters)
+            filters = Utils.parseFilter(JSONParser.parseObject(params.filters))
         }
 
         def results = TrackLog.findAll(
--- a/grails-app/controllers/io/xh/hoist/admin/TrackLogAdminController.groovy
+++ b/grails-app/controllers/io/xh/hoist/admin/TrackLogAdminController.groovy
@@ -9,13 +9,10 @@ package io.xh.hoist.admin
 
 import grails.gorm.transactions.ReadOnly
 import io.xh.hoist.BaseController
-import io.xh.hoist.data.filter.Utils
 import io.xh.hoist.security.Access
 import io.xh.hoist.track.TrackLog
 import io.xh.hoist.track.TrackLogAdminService
 
-import static io.xh.hoist.util.DateTimeUtils.*
-import static java.lang.Integer.parseInt
 
 @Access(['HOIST_ADMIN_READER'])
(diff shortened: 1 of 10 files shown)


grails-app/controllers/io/xh/hoist/admin/TrackLogAdminController.groovy
```diff
@@ -11,51 +11,25 @@ import grails.gorm.transactions.ReadOnly
...
-            if (params.browser)     browser =~ "%$params.browser%"
-            if (params.device)      device =~ "%$params.device%"
-            if (params.msg)         msg =~ "%$params.msg%"
-        }
-
-        renderJSON(results)
+        renderJSON(trackLogAdminService.queryTrackLog(parseRequestJSON()))
     }
 
     def lookups() {
```
lbwexler, line 27: lookups should come from the service as well

jacob-xhio: Resolved.

grails-app/services/io/xh/hoist/track/TrackLogAdminService.groovy
```diff
@@ -0,0 +1,65 @@
+package io.xh.hoist.track
```
amcclain, line 1: Pls add our standard copyright header.  (Does it actually have any magical powers or meaning? 🤷 but we do have it pretty consistently across the source code, and I think it remains a reasonable convention)

src/main/groovy/io/xh/hoist/data/filter/CompoundFilter.groovy
```diff
@@ -0,0 +1,54 @@
+/*
+ * This file belongs to Hoist, an application development toolkit
+ * developed by Extremely Heavy Industries (www.xh.io | info@xh.io)
+ *
+ * Copyright © 2023 Extremely Heavy Industries Inc.
```
amcclain, line 5: We can bump these to 2024 as well. It's how I personally keep track of what year it is. 😉

src/main/groovy/io/xh/hoist/data/filter/Utils.groovy
```diff
@@ -0,0 +1,146 @@
...
+ *
+ * Copyright © 2023 Extremely Heavy Industries Inc.
+ */
+
+package io.xh.hoist.data.filter
+
+import groovy.transform.CompileStatic
+
+@CompileStatic
+class Utils {
```
amcclain, line 13: Rename as `FilterUtils`?

src/main/groovy/io/xh/hoist/data/filter/Utils.groovy
```diff
@@ -0,0 +1,146 @@
...
+            }
+        }
+
+        throw new IllegalArgumentException("Unable to identify filter type: ${spec}")
+    }
+
+    /**
+     * Extract a list of fields used by filter
+     */
+    static List getFilterFields(Filter filter) {
```
amcclain, line 74: Return `Set<String>`?

A bit on the fence about letting function filters pass quietly through (vs throwing) - but hard to say what's best without knowing the use case.

amcclain (commented):
I know @lbwexler is reviewing so I took a quick look with some minor comments - I admit I didn't carefully look through the actual predicate construction logic as I didn't want to dupe up work, but happy to review more closely if helpful

amcclain:
Will also need changelog entry + determination if it's a breaking change to require https://github.com/xh/hoist-react/pull/3608

src/main/groovy/io/xh/hoist/data/filter/Utils.groovy
jacob-xhio: Will change return type to `Set` and throw runtime exception in function filter case

PR #344, opened 2024-04-02
Harmonize TrackLog and ClientError

jacob-xhio:
This PR adds new fields to TrackLog and ClientError databases to consolidate the two (to the extent that it makes sense)

It also refactors the `XhController` endpoints for `track` and `submitError` to expect the input parameters to be posted as JSON to the request body. One note about this change, because of the current implementation of the `ensureClientUsernameMatchesSession` method, `clientUsername` still needs to be posted to the server as a `param`. 

Finally, this server-side change refactors the `ClientErrorAdminController` to use a new service, the `ClientErrorAdminService`, to use filters provided by the client to do server-side filtering (this pattern is used in `TrackLogAdminController` and `TrackLogAdminService`).

### DIFF
--- a/grails-app/controllers/io/xh/hoist/impl/XhController.groovy
+++ b/grails-app/controllers/io/xh/hoist/impl/XhController.groovy
@@ -93,14 +93,14 @@ class XhController extends BaseController {
         ensureClientUsernameMatchesSession()
         def query = parseRequestJSON()
         trackService.track(
-            category: query.category? safeEncode(query.category.toString()): null,
-            msg: query.msg? safeEncode(query.msg.toString()): null,
-            data: query.data ? parseObjectOrArray(safeEncode(query.data.toString())) : null,
-            logData: query.logData? (query.logData == 'true' || query.logData == 'false' ? parseBoolean(query.logData.toString()) : query.logData.toString()?.split(',')): null,
-            elapsed: query.elapsed? query.elapsed: null,
-            severity: query.severity? safeEncode(query.severity.toString()): null,
-            url: query.url? safeEncode(query.url.toString()): null,
-            appVersion: query.appVersion? safeEncode(query.appVersion.toString()): null
+            category: safeEncode(query.category as String),
+            msg: safeEncode(query.msg as String),
+            data: parseObjectOrArray(safeEncode(query.data as String)),
+            logData: (query.logData == 'true' || query.logData == 'false') ? parseBoolean(query.logData as String) : (query.logData as String).split(','),
+            elapsed: query.elapsed,
+            severity: safeEncode(query.severity as String),
+            url: safeEncode(query.url as String),
+            appVersion: safeEncode(query.appVersion as String)
         )
         renderJSON(success: true)
     }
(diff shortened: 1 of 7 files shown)


grails-app/controllers/io/xh/hoist/impl/XhController.groovy
```diff
@@ -89,15 +89,18 @@ class XhController extends BaseController {
...
         ensureClientUsernameMatchesSession()
+        def query = parseRequestJSON()
         trackService.track(
-            category: safeEncode(category),
-            msg: safeEncode(msg),
-            data: data ? parseObjectOrArray(safeEncode(data)) : null,
-            logData: logData == 'true' || logData == 'false' ? parseBoolean(logData) : logData?.split(','),
-            elapsed: elapsed,
-            severity: safeEncode(severity)
+            category: query.category? safeEncode(query.category.toString()): null,
```
lbwexler, line 96: safeEnode is null safe,  so this could be written as `safeEnccode(query.category?.toString())`

But is the toString() really needed?  We want to avoid that boilerplate if possible

jacob-xhio: `toString` isn't needed. I had added that to stop intellij from yelling at me. Forgot to take it out

lbwexler: query.elapsed? is dangerous w.r.t. a value of zero.

grails-app/controllers/io/xh/hoist/impl/XhController.groovy
```diff
@@ -227,14 +230,15 @@ class XhController extends BaseController {
...
+    def submitError() {
         ensureClientUsernameMatchesSession()
+        def query = parseRequestJSON()
         clientErrorService.submit(
-            safeEncode(msg),
-            safeEncode(error),
-            safeEncode(appVersion),
-            safeEncode(url),
-            userAlerted
+            query.msg? safeEncode(query.msg.toString()): null,
```
lbwexler, line 237: see above

grails-app/controllers/io/xh/hoist/impl/XhController.groovy
jacob-xhio: Removed `query.elapsed?` ternary operator

lbwexler:
Looks great jacob -- will be nice to have these back in line

grails-app/controllers/io/xh/hoist/impl/XhController.groovy
```diff
@@ -89,15 +89,18 @@ class XhController extends BaseController {
...
         trackService.track(
-            category: safeEncode(category),
-            msg: safeEncode(msg),
-            data: data ? parseObjectOrArray(safeEncode(data)) : null,
-            logData: logData == 'true' || logData == 'false' ? parseBoolean(logData) : logData?.split(','),
-            elapsed: elapsed,
-            severity: safeEncode(severity)
+            category: safeEncode(query.category as String),
+            msg: safeEncode(query.msg as String),
+            data: query.data,
```
lbwexler, line 98: cleaner, but we lose the encoding now.

jacob-xhio: Resolved: passed `safeEncode: true` into `parseRequestJSON()`

PR #345, opened 2024-04-03
Fix NPE in DefaultRoleService.doLoadUsersForDirectoryGroups

ghsolomon:
Going to test this first in client-context

### DIFF
--- a/src/main/groovy/io/xh/hoist/role/provided/DefaultRoleService.groovy
+++ b/src/main/groovy/io/xh/hoist/role/provided/DefaultRoleService.groovy
@@ -190,9 +190,9 @@ class DefaultRoleService extends BaseRoleService {
         ldapService
             .lookupGroupMembers(foundGroups)
             .each {name, members ->
-                ret[name] = members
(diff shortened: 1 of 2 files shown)


src/main/groovy/io/xh/hoist/role/provided/DefaultRoleService.groovy
```diff
@@ -190,7 +190,9 @@ class DefaultRoleService extends BaseRoleService {
         ldapService
             .lookupGroupMembers(foundGroups)
             .each {name, members ->
-                ret[name] = members.collect(new HashSet()) { it.samaccountname.toLowerCase()}
+                ret[name] = members
+                    .collect(new HashSet()) { it.samaccountname?.toLowerCase() }
+                    .remove(null)
```
amcclain, line 195: My only suggestion would be to add a comment - something like `Exclude members without a samaccountname, e.g. email-only contacts within a DL`. 

I can imagine a year or two down the line, trying to puzzle out what case this was covering.

PR #347, opened 2024-04-03
Directory group error handling plus

lbwexler:
Still need to test this, but looking or feedback.  This AMs report at client highlights the importance of this.

### DIFF
--- a/grails-app/services/io/xh/hoist/ldap/LdapService.groovy
+++ b/grails-app/services/io/xh/hoist/ldap/LdapService.groovy
@@ -51,80 +51,106 @@ class LdapService extends BaseService {
 
 
     LdapPerson lookupUser(String sName) {
-        searchOne("(sAMAccountName=$sName) ", LdapPerson)
+        searchOne("(sAMAccountName=$sName) ", LdapPerson, true)
     }
 
     List<LdapPerson> lookupGroupMembers(String dn) {
-        // See class-level comment regarding this AD-specific query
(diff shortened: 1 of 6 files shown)


grails-app/services/io/xh/hoist/ldap/LdapService.groovy
```diff
@@ -51,83 +51,106 @@ class LdapService extends BaseService {
...
 
     LdapPerson lookupUser(String sName) {
-        searchOne("(sAMAccountName=$sName) ", LdapPerson)
+        searchOne("(sAMAccountName=$sName) ", LdapPerson, true)
     }
 
     List<LdapPerson> lookupGroupMembers(String dn) {
-        // See class-level comment regarding this AD-specific query
-        searchMany("(|(member0f=$dn) (memberOf:1.2.840.113556.1.4.1941:=$dn))", LdapPerson)
+        lookupGroupInternal(dn, true)
```
ghsolomon, line 58: Should this be `lookupGroupMembersInternal`?

ghsolomon:
Other than the comment re: `LdapService.lookupGroupMembers`, this looks good to me in that it will allow an app to get whatever users for directory groups are available upon first load, even if some LDAP servers are offline.

grails-app/services/io/xh/hoist/ldap/LdapService.groovy
lbwexler: good catch, thank you

lbwexler:
Cr: Greg

PR #354, opened 2024-04-25
Some new abstract and base classes for standardizing Oauth into Hoist

cnrudd:


amcclain (approved):
I added a few comments and a very minimal changelog entry in case we go looking for it ourselves in the future.  Good to merge as far as I am concerned, thanks!

PR #355, opened 2024-04-27
Code cleanups and smaller refactors to monitoring

amcclain:
+ Type `MonitorResult.metric` as `Number`
+ Rename MonitorsResults -> AggregateMonitorResult for clarity
+ New MonitorConfig class to formalize expected definition and types of required Hoist appConfig.
+ Apply `@GrailsCompileStatic` to services.
+ Additional doc comments.

### DIFF
--- a/grails-app/services/io/xh/hoist/monitor/MonitorReportService.groovy
+++ b/grails-app/services/io/xh/hoist/monitor/MonitorReportService.groovy
@@ -6,8 +6,10 @@
  */
 package io.xh.hoist.monitor
 
+import grails.compiler.GrailsCompileStatic
 import io.xh.hoist.BaseService
-import io.xh.hoist.util.Utils
+import io.xh.hoist.config.ConfigService
+import io.xh.hoist.email.EmailService
 
 import static io.xh.hoist.monitor.MonitorStatus.WARN
 import static io.xh.hoist.util.DateTimeUtils.MINUTES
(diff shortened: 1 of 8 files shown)


amcclain:
Sorry to add anything to your plate - threw a few ideas in here for discussion and to merge if you think they are moving in the right direction.

* Additional uses of `@GrailsCompileStatic` 
* MonitorResult.metric as Number vs Object - any sense apps were ever setting a non-numeric metric? Spurred by numeric comparisons in eval service w/tighter compilation settings.
* MonitorConfig object - happy to unwind but have been wondering for a while how we could better formalize the JSON configs and their types w/o lots of casting throughout the app. Realize this mints a new object each time but it's so lightweight hoping that's not material. 

Not at all married to these changes - could scale back and just take the rename and the doc comments which is what I really came here to do, but wanted to start reviewing HC code to see how we can tighten it down now that things have matured and these were some starting points I had in mind....

grails-app/services/io/xh/hoist/monitor/MonitorReportService.groovy
```diff
@@ -66,28 +77,14 @@ class MonitorReportService extends BaseService {
             emailService.sendEmail(
                 to: to,
                 subject: report.title,
-                html: formatHtml(report),
+                html: report.toHtml(),
```
lbwexler, line 80: nice!

lbwexler:
Love most everything.  thanks for the careful look.  

A little concerned that we are taking away the useful ability to have a non-numerical metric, could imagine that being useful. (e.g. a status code/string, reported by some system being tested?).  Will merge with that tiny change removed ( and perhaps we can discuss)

PR #357, opened 2024-05-02
Fixed DefaultRoleService completely failing if presented with a bad roleMember

jskupsik:
Currently, if a roleMember is present with roleMember.type==ROLE and an invalid/missing roleMember.name (one that does not match the Role.name of any Role), the DefaultRoleService will completely crash/fail, locking everyone out of the application.

This doesn't ever happen in normal operations/using the roles UI, but may happen if an application is manually changing roles/roleMembers (ie, migrating roles).

### DIFF
--- a/src/main/groovy/io/xh/hoist/role/provided/DefaultRoleService.groovy
+++ b/src/main/groovy/io/xh/hoist/role/provided/DefaultRoleService.groovy
@@ -346,8 +346,12 @@ class DefaultRoleService extends BaseRoleService {
                 if (!visitedRoles.contains(it)) {
                     visitedRoles << it
                     def effectiveRole = Role.get(it)
-                    rolesToVisit << effectiveRole
-                    ret << effectiveRole
+                    if (effectiveRole) {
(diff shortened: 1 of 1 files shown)


jskupsik:
### Test case: 
In my local toolbox env, I have added a row to the `xh_role_member` table with `type=ROLE`, `role_id=HOIST_ADMIN` (a role that exists), and `name=INVALID_ROLE_ID` (a role that does not exist). 

### Without fix:
RoleService completely breaks (no user has any roles loaded): 
Screenshot 2024-05-02 at 5 23 46 PM

The source of the issue is this cryptic error:
```2024-05-03 00:20:54.279 | 61f37439 | RoleService [ERROR] | Failure in timer | Cannot get property 'users' on null object [NullPointerException]```


### With fix:
App ignores the busted RoleMember rows and works, with this warn statement in the log:
```2024-05-03 00:18:59.822 | 3c5f0755 | RoleService [WARN] | Role HOIST_ADMIN references non-existent role INVALID_ROLE_ID | skipping```

PR #358, opened 2024-05-02
Fixed BaseService.initialize() hiding stack trace for init exceptions

jskupsik:
I have noticed that exceptions thrown by implementations of `BaseService.init()` using the `BaseService.parallelInit()` method do not show their full stack trace. This was because we were cutting off the exception "history" in `BaseService.initialize()`.

### DIFF
--- a/src/main/groovy/io/xh/hoist/BaseService.groovy
+++ b/src/main/groovy/io/xh/hoist/BaseService.groovy
@@ -28,6 +28,7 @@ import org.slf4j.LoggerFactory
 import org.springframework.beans.factory.DisposableBean
 
 import java.util.concurrent.ConcurrentHashMap
+import java.util.concurrent.ExecutionException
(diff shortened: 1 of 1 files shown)


lbwexler:
Jakub -- great catch, and this works, but I think there might be a more elegant way to do it, that avoids doubling up.  Just have a catch block first that catches `ExecutionException` and logs the `cause`

Sorry I missed this the first time around -- can't believe how long this has been wrong!!!!!

See
https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ExecutionException.html

Also See a similar bit in timer:
https://github.com/xh/hoist-core/blob/develop/src/main/groovy/io/xh/hoist/util/Timer.groovy
(here we also explicitly catch the timeout and rethrow a simpler exception, which you could also do, but not necessary)

PR #359, opened 2024-05-20
Fixed `defaultRoleService.list()` failing if presented with a bad roleMember

jskupsik:
Currently, if you attempt to access the roles page in an environment where there is a "bad roleMember", the `defaultRoleService.list()` fails with the message: `Cannot get property 'directoryGroups' on null object.`

Follow-up to https://github.com/xh/hoist-core/pull/357

### DIFF
--- a/grails-app/services/io/xh/hoist/role/provided/DefaultRoleAdminService.groovy
+++ b/grails-app/services/io/xh/hoist/role/provided/DefaultRoleAdminService.groovy
@@ -39,7 +39,7 @@ class DefaultRoleAdminService extends BaseService {
             errorsForGroups = emptyMap()
 
         if (defaultRoleService.directoryGroupsSupported) {
-            Set<String> groups = roles.collectMany(new HashSet()) { it.directoryGroups }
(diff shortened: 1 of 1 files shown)


grails-app/services/io/xh/hoist/role/provided/DefaultRoleAdminService.groovy
```diff
@@ -39,7 +39,7 @@ class DefaultRoleAdminService extends BaseService {
             errorsForGroups = emptyMap()
 
         if (defaultRoleService.directoryGroupsSupported) {
-            Set<String> groups = roles.collectMany(new HashSet()) { it.directoryGroups }
+            Set<String> groups = roles.collectMany(new HashSet()) { it?.directoryGroups }
```
lbwexler, line 42: sorry, I don't get this -- Role.list() can return nulls?!?  I don't think I have ever encountered this.

lbwexler (commented):
Want to understand how we are getting these "Bad Roles" before merging.

PR #360, opened 2024-05-22
BaseUserService now allows app implementation to use the app's User type

jskupsik:
Currently, BaseUserService does not allow the implementation to use it's own User:
```groovy
class AppUser extends HoistUser { ... 
```
```groovy
class UserService extends BaseUserService {
    // This does not build, because the return type of `List<AppUser> ` is incompatible with `List<HoistUser>`
    @Override
    List<AppUser> list(boolean activeOnly) { ... 
```
With this fix, the app's UserService is allowed to type its own methods.


jskupsik:
Toolbox branch, which shows an implementation that now works due to this fix - https://github.com/xh/toolbox/pull/720

PR #361, opened 2024-06-03
Enhancements for BaseProxyService

saba-mo:


grails-app/services/io/xh/hoist/http/BaseProxyService.groovy
```diff
@@ -121,9 +126,9 @@ abstract class BaseProxyService extends BaseService {
     }
 
     protected CloseableHttpClient getSourceClient() {
-        def ret = _sourceClient
```
lbwexler, line 124: I am confused -- how was this ever working/caching?  I don't see  _sourceClient getting set.

grails-app/services/io/xh/hoist/http/BaseProxyService.groovy
```diff
@@ -75,7 +76,11 @@ abstract class BaseProxyService extends BaseService {
         try (CloseableHttpResponse sourceResponse = sourceClient.execute(method)) {
             response.setStatus(sourceResponse.code)
             installResponseHeaders(response, sourceResponse)
-            sourceResponse.entity.writeTo(response.outputStream)
+
+            if (sourceResponse.entity) {
+                sourceResponse.entity.writeTo(response.outputStream)
```
lbwexler, line 81: this is great.  Please note could also be replaced with 

`sourceResponse.entity?.writeTo(response.outputStream)`

grails-app/services/io/xh/hoist/http/BaseProxyService.groovy
saba-mo: I was wondering the same thing...maybe it would have to get set in the app to actually be cached?

PR #368, opened 2024-07-03
Remove compileStatic annotation.

cnrudd:
Remove compileStatic annotation.
It was preventing the `keys` call in the LdapObject `populate` method from using the subclass' `getKeys` method.

CompileStatic was added recently in this commit https://github.com/xh/hoist-core/pull/364/commits/2c85e6ab644d95f53b91d6cd490aa76ff3cac7ad merged in this PR: https://github.com/xh/hoist-core/pull/364

The bug was found by a client using the `findGroups` method. He noticed that the `member` key on the retuned`LdapGroup` result was always `null`.  The `findGroups` method is not used in any production code, so this does not require an urgent release.

This does not affect the  `DefaultRoleService` because it uses the `lookupGroupMembers` method.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@
 
 * Remove obsolete, non-functioning GSP support from EmailService.
 
+### 🐞 Bug Fixes
+
+* `compileStatic` annotation removed from Ldap objects.  It was preventing the `keys` call in the
+  LdapObject `populate` method from using the subclass' `getKeys` method.
+
 ## 20.2.0 - 2024-06-26
 
 ### ⚙️ Technical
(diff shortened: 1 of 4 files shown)


lbwexler:
We don't 100% understand this, and it points to something we are doing wrong with static methods in java.  

Reverting the change to compilestatic is pragmatic for the quick fix.

jskupsik:
I think we ran into the difference between "overriding" normal methods and "hiding" static methods in Java: https://docs.oracle.com/javase/tutorial/java/IandI/override.html

> The version of the hidden static method that gets invoked depends on whether it is invoked from the superclass or the subclass.

amcclain:
I had the same thought - that this is likely some kind of code smell here and that there might be a better way to be handling this. I'd like to start using `@CompileStatic` more, as it does catch several entire classes of bugs that we hit somewhat regularly on our server-side code. 

But obviously we need to understand it and how to detect + properly workaround any behavior changes or restrictions it puts in place. Apologies for causing some trouble with the addition on this class, makes sense to pull it back until we can revisit. 👍

PR #369, opened 2024-07-12
Fix configuration of Hazelcast distributed objects

lbwexler:
Should be a simplification and more straightforward use of the hazelcast config API.  The get methods already do clone the current configuration and "add" a new configuration.  I think we may have been thrown off with a very specific bug around the failure to deep clone evictionConfig

### DIFF
--- a/grails-app/domain/io/xh/hoist/role/provided/Role.groovy
+++ b/grails-app/domain/io/xh/hoist/role/provided/Role.groovy
@@ -33,12 +33,13 @@ class Role implements JSONFormat {
         lastUpdatedBy maxSize: 50
     }
 
+    // We don't expect a huge number of roles.
+    // Tighten as examples of custom cache and collection query size
     static cache = {
-        evictionConfig.size = 5000
+        evictionConfig.size = 1000
     }
-
     static membersCollectionCache = {
-        evictionConfig.size = 5000
+        evictionConfig.size = 1000
     }
 
     static beforeInsert = {
--- a/grails-app/domain/io/xh/hoist/role/provided/Role.groovy
+++ b/grails-app/domain/io/xh/hoist/role/provided/Role.groovy
@@ -34,11 +34,8 @@ class Role implements JSONFormat {
     }
 
     // We don't expect a huge number of roles.
-    // Tighten as examples of custom cache and collection query size
+    // Tighten as example of a custom cache config being applied to a collection cache (Role.members)
     static cache = {
-        evictionConfig.size = 1000
-    }
-    static membersCollectionCache = {
         evictionConfig.size = 1000
     }
 
--- a/grails-app/domain/io/xh/hoist/role/provided/Role.groovy
+++ b/grails-app/domain/io/xh/hoist/role/provided/Role.groovy
@@ -33,6 +33,14 @@ class Role implements JSONFormat {
         lastUpdatedBy maxSize: 50
     }
 
+    static cache = {
+        evictionConfig.size = 5000
+    }
+
+    static membersCollectionCache = {
+        evictionConfig.size = 5000
+    }
+
     static beforeInsert = {
         if (Role.findByNameIlike(name)) {
             throw new RoutineRuntimeException('Role Name must be case-insensitive unique.')
--- a/grails-app/init/io/xh/hoist/ClusterConfig.groovy
+++ b/grails-app/init/io/xh/hoist/ClusterConfig.groovy
@@ -168,7 +168,7 @@ class ClusterConfig {
                 // workaround - hz does not clone evictionConfig
                 baseConfig.evictionConfig = new EvictionConfig(baseConfig.evictionConfig)
                 customizer.delegate = baseConfig
-                customizer.resolveStrategy = DELEGATE_FIRST
+                customizer.resolveStrategy = Closure.DELEGATE_FIRST
                 customizer(baseConfig)
             }
         }
@@ -198,7 +198,7 @@ class ClusterConfig {
                         throw new RuntimeException('Unable to configure Cluster object')
                 }
                 customizer.delegate = objConfig
-                customizer.resolveStrategy = DELEGATE_FIRST
+                customizer.resolveStrategy = Closure.DELEGATE_FIRST
                 customizer(objConfig)
             }
         }
--- a/grails-app/init/io/xh/hoist/ClusterConfig.groovy
+++ b/grails-app/init/io/xh/hoist/ClusterConfig.groovy
@@ -1,22 +1,16 @@
 package io.xh.hoist
 
-import com.hazelcast.config.CacheSimpleConfig
 import com.hazelcast.config.Config
 import com.hazelcast.config.EvictionConfig
 import com.hazelcast.config.EvictionPolicy
 import com.hazelcast.config.InMemoryFormat
-import com.hazelcast.config.MapConfig
-import com.hazelcast.config.ReplicatedMapConfig
-import com.hazelcast.config.SetConfig
-import com.hazelcast.config.TopicConfig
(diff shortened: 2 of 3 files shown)


jskupsik (approved):
Tested locally - the evicitionConfig.size property now successfully changes as specified.

grails-app/init/io/xh/hoist/ClusterConfig.groovy
```diff
@@ -163,14 +163,27 @@ class ClusterConfig {
     private void createHibernateConfigs(Config config) {
         grailsApplication.domainClasses.each { GrailsClass gc ->
             Closure customizer = gc.getPropertyValue('cache') as Closure
-            def baseConfig = config.getCacheConfig(gc.fullName)
             if (customizer) {
+                def baseConfig = config.getCacheConfig(gc.fullName)
```
lbwexler, line 167: I was intentionally getting the cach out of the loop -- found otherwise, the uncustomized values would *not* get the default, as if something later on in the hazelcast lifecycle would apply a different value.    I should have commented, as without that fact, your change makes 100% sense

grails-app/domain/io/xh/hoist/role/provided/Role.groovy
```diff
@@ -33,6 +33,12 @@ class Role implements JSONFormat {
         lastUpdatedBy maxSize: 50
     }
 
+    // We don't expect a huge number of roles.
+    // Tighten as example of a custom cache config being applied to a collection cache (Role.members)
```
amcclain, line 37: Trying to understand the second line of this comment - how does the config below get applied to the `members` collection, vs the cache of Role objects themselves (as the first line indicates)?

Could this comment have referred to a previous customization that was then removed?

grails-app/init/io/xh/hoist/ClusterConfig.groovy
```diff
@@ -165,17 +162,25 @@ class ClusterConfig {
...
+            // Pre-access cache for all domain classes to ensure we capture the common 'default'
+            // (Not clear why this is needed -- but hibernate would otherwise create these differently)
+            def configs = [
+                // 1) Main 2nd-level entity cache
+                config.getCacheConfig(gc.fullName),
+                // 2) Collection caches
+                config.getCacheConfig(gc.fullName + '.*')
+            ]
+
+            // apply any app customization
```
amcclain, line 174: Is this replicating logic that Grails previously / otherwise handled, but we need to take on here given our use of Hazelcast?  Would like to capture a bit more context here about why this is needed while it's still in your heads - reading through without that context, I'm not sure if this is all in place due to a) general use of Hazelcast as alt cache provider, b) HZ bugs or limitations, c) Grails/GORM bugs or limitations, or something else....

grails-app/init/io/xh/hoist/ClusterConfig.groovy
```diff
@@ -165,17 +162,25 @@ class ClusterConfig {
...
-                cacheConfig.evictionConfig = new EvictionConfig(baseConfig.evictionConfig)
-                customizer.delegate = cacheConfig
-                customizer.resolveStrategy = Closure.DELEGATE_FIRST
-                customizer(cacheConfig)
+                configs.each { cfg ->
+                    // workaround - hz does not clone evictionConfig
+                    cfg.evictionConfig = new EvictionConfig(cfg.evictionConfig)
+                    customizer.delegate = cfg
+                    customizer.resolveStrategy = Closure.DELEGATE_FIRST
+                    customizer(cfg)
```
amcclain, line 182: These lines the most opaque to me - would not have figured this out, for sure!  Realize it is not new, but is there any extra pointer / link / explanation we could offer in a comment here?  Especially the bit about re-pointing the `customizer` closure to one config and then the next and calling it with both - just trying to build a model in my head of what that does.

grails-app/init/io/xh/hoist/ClusterConfig.groovy
```diff
@@ -187,20 +192,18 @@ class ClusterConfig {
             objs.forEach {String key, List value ->
                 def customizer = value[1] as Closure,
                     objConfig
-                // IMPORTANT -- We do an explicit clone, because wild card matching in hazelcast will
-                // actually just return the *shared* config (?!), and never want to let app edit that.
                 switch (value[0]) {
                     case IMap:
-                        objConfig = new MapConfig(config.findMapConfig(gc.fullName + '_' + key))
+                        objConfig = config.getMapConfig(gc.fullName + '_' + key)
```
amcclain, line 197: Did this come down to a difference between `findMapConfig` and `getMapConfig`, where the latter turned out to be what we wanted/needed, without need for clone?

grails-app/domain/io/xh/hoist/role/provided/Role.groovy
```diff
@@ -33,6 +33,12 @@ class Role implements JSONFormat {
...
+    static cache = {
+        evictionConfig.size = 1000
```
amcclain, line 39: To ensure I'm understanding - this is how we specify the max number of cached object instances we want to support for this domain object?  

Looks like the default is 5000, set below, is that correct? Does that actually reserve memory somehow for the max allowed number of elements?  I.e. what benefit do we get from making this value explicitly lower?

amcclain (commented):
Thank you so much guys for getting down into it with these caching configs!

A few questions / comments I had - mostly wondering if there is any additional context we could add to the existing comments, to help guide future-us and other app developers around these customizations.

grails-app/domain/io/xh/hoist/role/provided/Role.groovy
amcclain: Looked again at this and realizing - is this `static cache = {}` closure on domain objects  a new property that we support through the custom code below in ClusterConfig?  I.e. not supported by Grails/GORM but a Hoist feature?

lbwexler:
yes

grails-app/domain/io/xh/hoist/role/provided/Role.groovy
lbwexler: Yes, this is our own construct, not new.  Setting this to a *lower* value, because it seemed safe enough and was a built-in example of customizing the cache in a safe way, but yes, arbitrary -- we would not expect apps to typically shrink this value

grails-app/domain/io/xh/hoist/role/provided/Role.groovy
lbwexler: We do apply it to *both* -- comment and code should be clearer about that

grails-app/init/io/xh/hoist/ClusterConfig.groovy
lbwexler: Previously we could specify this in an `ehcache.xml` file which was not heavily used or commented, but available.  We were looking to replace that functionality with hazelcast, and this is what we came up with.  

Note that we folded this in to a general system for customizing Hazelcast caches.    Also note that this really had to happen as a static code block that could be executed by our framework very early on in the lifecycle.  Once grails "initializes" hibernate with the apps Domain objects these caches are created, and any customization is too late.

grails-app/init/io/xh/hoist/ClusterConfig.groovy
lbwexler: Agree -- this was a late change and we should comment it, but also be sure its correct.

We are assuming that apps will be ok to make the very same customization to the CollectionCaches for a Hibernate object as the core entity cache.  It seemed like a reasonable assumption, given -- as Jakub astutely pointed out -- that these caches seem to track identically in terms of their size.   

My feeling was that if we ever needed to provide an alternative closure to customize a particular CollectionCache we could.

PR #371, opened 2024-07-17
Simplified the xhDbConnectionMonitor query to work with more SQL dialects

jskupsik:
The existing query, `SELECT * FROM xh_monitor WHERE code = 'xhDbConnectionMonitor' LIMIT 1`, does not work in SQL Server (Microsoft SQL).
The equivalent statement in SQL Server would be `SELECT TOP(1) * FROM xh_monitor WHERE code = 'xhDbConnectionMonitor'`

I propose we simplify it to the SQL statement `SELECT * FROM xh_monitor WHERE code = 'xhDbConnectionMonitor'`, which should work in all dialects.

### DIFF
--- a/src/main/groovy/io/xh/hoist/monitor/provided/DefaultMonitorDefinitionService.groovy
+++ b/src/main/groovy/io/xh/hoist/monitor/provided/DefaultMonitorDefinitionService.groovy
@@ -116,7 +116,7 @@ class DefaultMonitorDefinitionService extends BaseService {
         def startTime = currentTimeMillis()
         Sql sql = new Sql(dataSource)
         try {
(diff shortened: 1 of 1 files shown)


jskupsik:
We could also potentially do `SELECT 1 FROM xh_monitor WHERE code = 'xhDbConnectionMonitor'` - if we are worried about `*` fetching too much data.

amcclain:
Not sure why we had that LIMIT 1 in there given that code has a unique constraint - thanks for fixing!

amcclain:
Wish we had just made this configurable via params - found it failing on a deployment, went to adjust the monitor config then remembered this is hard-coded.

lbwexler:
I don't feel too bad about that -- I think this is more appropriate than config driven sql would have been

PR #373, opened 2024-07-23
Bulk role category update

Ryanseanlee:


grails-app/services/io/xh/hoist/role/provided/DefaultRoleUpdateService.groovy
```diff
@@ -106,6 +106,27 @@ class DefaultRoleUpdateService extends BaseService {
...
+        ensureEnabled()
+        List<Role> updatedRoles = new ArrayList<Role>()
+        roles.each { roleName ->
+            Role role = Role.get(roleName as String)
+            role.category = category
+            role.save(flush:true)
+            updatedRoles.push(role)
+        }
+        trackService.track(
+            msg: "Updated '${roles.size()}' roles.",
```
amcclain, line 120: Nit - no need for quotes around the count here.  Track statement looks great otherwise

grails-app/services/io/xh/hoist/role/provided/DefaultRoleUpdateService.groovy
```diff
@@ -106,6 +106,27 @@ class DefaultRoleUpdateService extends BaseService {
         }
     }
 
+    @Transactional
+    List<Role> bulkCategoryUpdate(ArrayList<String> roles, String category) {
```
amcclain, line 110: Would use the more generic `List` in the method signature here for incoming role names

grails-app/services/io/xh/hoist/role/provided/DefaultRoleUpdateService.groovy
```diff
@@ -106,6 +106,27 @@ class DefaultRoleUpdateService extends BaseService {
...
+        ensureEnabled()
+        List<Role> updatedRoles = new ArrayList<Role>()
+        roles.each { roleName ->
+            Role role = Role.get(roleName as String)
```
amcclain, line 114: and shouldn't need the `as String` cast here

grails-app/services/io/xh/hoist/role/provided/DefaultRoleUpdateService.groovy
```diff
@@ -106,6 +106,27 @@ class DefaultRoleUpdateService extends BaseService {
...
+        }
+        trackService.track(
+            msg: "Updated '${roles.size()}' roles.",
+            category: 'Audit',
+            data: [
+                roles                 : roles,
+                category              : category
+            ]
+        )
+        defaultRoleService.refreshRoleAssignments()
```
amcclain, line 127: This is a special case where I don't think we need/want to call `refreshRoleAssignments` - this is dedicated to changing the category only, which won't affect the assignments or how they are reported (who belongs to what role).  I would leave it off, mainly to avoid the server doing extra work it doesn't need to do.

Only counter-argument would be if we were concerned the scope of what this method does could change in the future and we want to always ensure we have this call in place when we touch anything about roles.

amcclain (commented):
Looking good - some minor comments to consider below. Let's also add a changelog entry before merging.

lbwexler:
Am going to merge now -- will take care of ATMs comments in a subsequent committ

PR #375, opened 2024-07-24
Monitor exception as string

lbwexler:


amcclain (approved):
Thanks for the fix - it looks good to me. (Although to be clear I did _not_ run this, hence me not merging myself.)

I wondered if `summaryTextForThrowable` output would require any adjustments on the admin client display, but I see our comment on it is "The default implementation is designed to yield meaningful information within a one-line summary." which sounds like what we want for this use case.

PR #376, opened 2024-07-26
Support deleting roles with dots or other special chars in their name

cnrudd:
Fixes issue https://github.com/xh/hoist-core/issues/370

Breaking change due to corresponding required change in [hoist-react](https://github.com/xh/hoist-react/pull/3735)

Grails' dot splitting appears to be designed to handle "Optional File Extensions" (https://docs.grails.org/latest/guide/theWebLayer.html#embeddedVariables)
If we adjusted our grails mapping and controller to "put the value back together" with `response.format`, it would likely avoid needing a corresponding change in hoist-react, and probably support names with one dot, but it would be hacking over a mis-use of the grails conventions, I think.


ghsolomon (approved):

grails-app/controllers/io/xh/hoist/admin/RoleAdminController.groovy
```diff
@@ -43,9 +43,10 @@ class RoleAdminController extends BaseController {
         renderJSON(data: role)
     }
 
-    def delete(String id) {
+    def delete() {
         ensureHoistRoleManager()
-        defaultRoleUpdateService.delete(id)
+        Map roleSpec = parseRequestJSON()
+        defaultRoleUpdateService.delete(roleSpec.name)
```
amcclain, line 49: We might want to make this `role` vs. `name` - for symmetry for bulk update params in #373 where we have `roles` as key in request JSON.

amcclain: Eh - happy to leave as is - this is more consistent with single-role create / update, with JSON body with `name` key

amcclain (approved):

PR #378, opened 2024-07-29
Improve `EmailService` subjects generated for non-prod / overridden deliveries

lbwexler:
Went way too deep down this rabbit hole -- but there was nothing else to do but finish it

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## 21.0-SNAPSHOT - unreleased
 
+### ⚙️ Technical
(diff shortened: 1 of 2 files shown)


amcclain:
Pushed a few additional tweaks + internal typings while reviewing - I read through it and looks like a nice cleanup, but _did not run or test myself_ .

@jskupsik if you have a chance over the next day or two, can you please review and test out the expected use cases so we can safely get this in?  Many thanks.

jskupsik (approved):

PR #379, opened 2024-07-31
Report if `xhLdapConfig.enabled = true` + `xhLdapUsername = 'none'`

Ryanseanlee:


amcclain (approved):
Sorry for my thrashing on this - I believe this covers the one case I wanted to handle better - `xhLdapConfig.enabled`  set to true, but separate `xhLdapUsername` left on default value of `'none'`.

Yes, this would not catch any other invalid value - presumably the ldap bind woudl throw its own exception - but given that we create and read these two configs of ours, a little additional clarity for developer if they are mismatched seems worth it.

lbwexler:
This is a really good change -- thanks for seeing it through 👍🏻

PR #380, opened 2024-08-05
Clarify `EnvironmentService` handling of invalid `xhAppTimeZone` config

minusleaf:
Fixes the issue where the log would never call because `configZoneId` given a bad `xhAppTimeZone` value would return the `defaultZone`. This would then be considered valid as an available ID by the conditional and therefore not log if we use the default or not.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,11 +2,6 @@
 
 ## 21.0-SNAPSHOT - unreleased
 
-### 🐞 Bug Fixes
-
-* Fix in `EnvironmentService` to show error log when passed in an invalid time zone ID by setting to
-  default time zone after the check.
-
 ## 20.4.0 - 2024-07-31
 
 ### 🐞 Bug Fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## 21.0-SNAPSHOT - unreleased
 
+### 🐞 Bug Fixes
+
+* Fix in `EnvironmentService` to show error log when passed in an invalid time zone ID by setting to
+  default time zone after the check.
+
 ## 20.4.0 - 2024-07-31
 
 ### 🐞 Bug Fixes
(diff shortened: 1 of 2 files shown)


minusleaf:
Need to fix the return

grails-app/services/io/xh/hoist/environment/EnvironmentService.groovy
```diff
@@ -89,10 +89,11 @@ class EnvironmentService extends BaseService {
     //---------------------
     private TimeZone calcAppTimeZone() {
         def defaultZone = 'UTC',
-            configZoneId = configService.getString('xhAppTimeZone', defaultZone)
+            configZoneId = configService.getString('xhAppTimeZone')
```
amcclain, line 92: It's OK (and good) to remove this default here, but note that's only because we always ensure that the `xhAppTimeZone` config is created - that happens in Bootstrap.groovy.  Before we had that code, this second arg was important as otherwise configService would throw if the config was not found in the DB.

Given the bootstrap routine, we know the config will always be there, but we now are defending against it being some unexpected/unknown zone.

As we discussed (and for the record here), `TimeZone.getTimeZone('someBadZoneString')` actually does default to `GMT`, which is functionally the same as `UTC` although not identical in Java world (different actual `TimeZone` objects, same offset).

So the code we had before would work, but it was a bit odd and did not get you exactly what it promised, so I'm still happy to have our improvement here.

CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 ## 21.0-SNAPSHOT - unreleased
 
+### 🐞 Bug Fixes
+
+* Fix in `EnvironmentService` to show error log when passed in an invalid time zone ID by setting to
+  default time zone after the check.
```
amcclain, line 8: This isn't quite correct - we were always printing that error log, the code now is just a bit more explicit and exact in terms of how it is falling back to the default.

But TBH I don't think we need a CL entry here - this is a very minor code cleanup, not really a bug, and no functional difference. Pls remove it and then happy to merge. Thanks!

PR #386, opened 2024-08-28
CachedValue and related enhancements to multi-instance support

lbwexler:
Needed to remove the `oldValue` in our change handler -- we could bring it back optionally, but it will require giving up on our performance enhancements to avoid serializing "old" versions of our large data sets


src/main/groovy/io/xh/hoist/cache/CachedValue.groovy
```diff
@@ -0,0 +1,160 @@
...
+            return null
+        }
+        return ret?.value as V
+    }
+
+    /** Set the value. */
+    void set(V value) {
+        def oldEntry = _map[name]
+        oldEntry?.isRemoving = true
+        _map[name] = new Entry(name, value, svc.instanceLog.name)
```
TomTirapani, line 83: If value is null, should we set to null here rather than an Entry with null value? Thinking about symmetry with Cache's remove()
e.g. `_map[name] = value == null ? null : new Entry(name, value, svc.instanceLog.name)`

TomTirapani (approved):

src/main/groovy/io/xh/hoist/cache/CachedValue.groovy
lbwexler: good catch!  Will do

amcclain:
Will look with care tomorrow - eager to see it, thanks for setting this up.

amcclain:
Please see proposed updates in #388

src/main/groovy/io/xh/hoist/cache/BaseCache.groovy
```diff
@@ -0,0 +1,110 @@
...
+
+    /** Closure to determine the timestamp of an entry. (optional) */
+    public final Closure timestampFn
+
+    /** Whether this cache should be replicated across a cluster. Default false */
+    public final boolean replicate
+
+    /**
+     * Optimize removals of replicated entries, such that the old value is
+     * not re-serialized. Default false.
```
TomTirapani, line 41: Thanks for adding this back in!

That said, I wonder if the default should be `optimizeRemovals: true`?

I imagine **most** caches will not add change handlers, and **most** change handlers will not be interested in the old value, so why not make it more efficient by default and have the `oldValue` something you have to opt-in to? Or perhaps `isOptimizedRemoval` is always true if you don't have any change handlers, regardless of this flag?

src/main/groovy/io/xh/hoist/cache/BaseCache.groovy
```diff
@@ -0,0 +1,110 @@
...
+     *
+     * This optimization is useful for caches containing large objects requiring
+     * time to to serialize/deserialize.  NOTE -- if enabled, the CacheValueChanged
+     * events fired on this object will *not* contain the oldValue.
+     */
+    public final boolean optimizeRemovals
+
+    protected final List<Closure> onChange = []
+
+    BaseCache(Map options) {
```
TomTirapani, line 51: Any reason not to include onChange handlers in the constructor? Seems a bit strange that you have to add them as second step

src/main/groovy/io/xh/hoist/cache/BaseCache.groovy
amcclain: Good point! Should we make it more specific and opt-in? Eg onChangeIncludesOldValue or something along those lines?

lbwexler:
CR: Tom, ATM, Greg

PR #387, opened 2024-08-28
Update `/xh` endpoints and config for version/instance change detection

ghsolomon:
…ropriate defaults


ghsolomon:
Associated with HR PR: https://github.com/xh/hoist-react/pull/3768

grails-app/controllers/io/xh/hoist/impl/XhController.groovy
```diff
@@ -219,8 +219,8 @@ class XhController extends BaseController {
         renderJSON(environmentService.getEnvironment())
     }
 
-    def version() {
-        def options = configService.getMap('xhAppVersionCheck', [:])
+    def status() {
```
lbwexler, line 222: confusing with our status monitors?

lbwexler: connectionCheck? heartbeat?

grails-app/controllers/io/xh/hoist/impl/XhController.groovy
```diff
@@ -219,8 +219,8 @@ class XhController extends BaseController {
         renderJSON(environmentService.getEnvironment())
     }
 
-    def version() {
```
amcclain, line 222: I would like to avoid removing this entirely - at least not right away.  It might well be used for liveness checks for k8s and similar - I am not positive about that, but I think it's likely enough to warrant keeping it around.

All in favor of adding the new endpoint, though - `/version` could be slimmed down to truly just the version/build. I do _not_ think that any potential non-Hoist integrations care about the payload, only that it will respond.

grails-app/controllers/io/xh/hoist/impl/XhController.groovy
amcclain: * We could continue with `/status`, but plan to expand the payload here at a later point to include 
 health/readiness information. 
  * Could include some highly summarized version of the actual status monitor outputs (eg counts only), if only to acknowledge those have some claim to "status". Although for this whitelisted endpoint we obvs would not want to output anything more detailed (not even, say, the names of monitors).
  * But thinking ahead to this including some general `ready:true` indicator helps to justify the name - which I like (mostly because it's difficult to come up with another, better name...)
* `/instance`, `/instanceInfo` or `/instanceDetails` would be my other suggestion. It might be a bit weird to have the other version-check related instructions in a return with that name, but not _that_ weird.  
  * If we felt the need, we could always nest those under a sub-key in the return, like `versionCheck`

tldr; I am OK with `status` if we plan to expand the payload to ready checks, but suggest `instance` otherwise.

grails-app/init/io/xh/hoist/BootStrap.groovy
```diff
@@ -99,17 +99,17 @@ class BootStrap implements LogSupport {
                 groupName: 'xh.io',
                 note: 'Official TimeZone for this application - e.g. the zone of the head office. Used to format/parse business related dates that need to be considered and displayed consistently at all locations. Set to a valid Java TimeZone ID.'
             ],
-            xhAppVersionCheck: [
+            xhAppStatusCheck: [
```
amcclain, line 102: This config could then either become `xhAppStatusCheck` or `xhInstanceCheck`.  

Neither really get at the fact that these are instructions for the TS client, but then again, what else would they be for?

grails-app/init/io/xh/hoist/BootStrap.groovy
```diff
@@ -99,17 +99,17 @@ class BootStrap implements LogSupport {
...
                 valueType: 'json',
                 defaultValue: [
-                    interval: configService.getInt('xhAppVersionCheckSecs', 30),
-                    mode: configService.getBool('xhAppVersionCheckEnabled', true) ? 'promptReload' : 'silent'
+                    interval: 10,
+                    updateMode: configService.getMap('xhAppVersionCheck', [mode: 'promptReload']).get('mode')
```
amcclain, line 106: Forgot we had this pattern in place with bootstrap to read values from the previous/deprecated config - great!

amcclain: Clearer to make this `onVersionChange`?  Just to make it clear that we are talking about the version updating, and not the instance.

CHANGELOG.md
```diff
@@ -6,9 +6,16 @@
...
 * Requires `hoist-react >= 67.0.0` for client-side changes to accommodate updated `track`
   and `submitError` APIs. See below for database column additions to support the same.
-  
+
 * Requires `hoist-react >= 67.0` to use corresponding role delete bug fix.
 
+* Deprecated config `xhAppVersionCheck` in favor of new `xhAppStatusCheck`. `Bootstrap` will
+  migrate apps' existing `mode` value to this new config's `updateMode` property. Since status
+  checks now include `instanceName`, `interval` should be set to a tighter value than before.
+  Default is 10s. Requires `hoist-react >= 67.0`.
```
amcclain, line 15: We can indicate "Requires `hoist-react >= 67.0`" once at the top of the entry - quite clear now users will be upgrading both ;)

amcclain:
Had a discussion with Lee earler this evening and throwing these changes over the wall to see if you guys think this holds together.  

While the client-side code does not save down all the data produced by `/xh/environment` on its polling checks, I was still happy to not need another endpoint and this should always be cheap / safe to call.  

Retained `/xh/version` for compatibility, but delegating to the same.  Was also looking at what whitelisted `/xh/environment` has been returning to unauthenticated clients and felt like it was more than we would actually want - especially the library versions. But do not believe the slimmed down payload would break any current usages.

amcclain (approved):
Looks like a good place to land to me 👍

PR #388, opened 2024-08-30
Proposed usage of @NamedVariant for Cache classes

amcclain:
+ Minor updates to ctor validations
+ Doc comment tweaks
+ Set `@CompileStatic` on AlertBannerService for good measure

### DIFF
--- a/src/main/groovy/io/xh/hoist/cache/Cache.groovy
+++ b/src/main/groovy/io/xh/hoist/cache/Cache.groovy
@@ -9,7 +9,9 @@ package io.xh.hoist.cache
 
 import com.hazelcast.replicatedmap.ReplicatedMap
 import groovy.transform.CompileStatic
-
+import groovy.transform.NamedParam
+import groovy.transform.NamedVariant
+import io.xh.hoist.BaseService
 import io.xh.hoist.util.Timer
 
 import java.util.concurrent.ConcurrentHashMap
@@ -27,15 +29,27 @@ class Cache<K,V> extends BaseCache<V> {
     private final Map<K, Entry<V>> _map
     private final Timer timer
 
-    Cache(Map options) {
-        super(options)
+    @NamedVariant
+    Cache(
+        @NamedParam(required = true) BaseService svc,
+        @NamedParam String name,
+        @NamedParam Object expireTime = null,
+        @NamedParam Closure expireFn = null,
+        @NamedParam Closure timestampFn = null,
+        @NamedParam boolean replicate = false,
+        @NamedParam boolean optimizeRemovals = false
+    ) {
+        super(svc, name, expireTime, expireFn, timestampFn, replicate, optimizeRemovals)
+
         if (replicate && !name) {
-            throw new RuntimeException("Cannot create a replicated cache without a unique name")
+            throw new IllegalArgumentException("Cannot create a replicated Cache without a unique name")
         }
 
-        _map = replicate ? svc.getReplicatedMap(name) : new ConcurrentHashMap()
-        if (replicate) {
-            (_map as ReplicatedMap).addEntryListener(getHzEntryListener())
+        // Intentional `this` below - super might override to false if !multiInstanceEnabled
+        _map = this.replicate ? svc.getReplicatedMap(name) : new ConcurrentHashMap()
+
+        if (_map instanceof ReplicatedMap) {
+            _map.addEntryListener(getHzEntryListener())
         }
 
         timer = new Timer(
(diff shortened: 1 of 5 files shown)


src/main/groovy/io/xh/hoist/cache/Cache.groovy
```diff
@@ -27,15 +29,27 @@ class Cache<K,V> extends BaseCache<V> {
...
         if (replicate && !name) {
-            throw new RuntimeException("Cannot create a replicated cache without a unique name")
+            throw new IllegalArgumentException("Cannot create a replicated Cache without a unique name")
         }
 
-        _map = replicate ? svc.getReplicatedMap(name) : new ConcurrentHashMap()
-        if (replicate) {
-            (_map as ReplicatedMap).addEntryListener(getHzEntryListener())
+        // Intentional `this` below - super might override to false if !multiInstanceEnabled
+        _map = this.replicate ? svc.getReplicatedMap(name) : new ConcurrentHashMap()
```
amcclain, line 49: This `this` is a bit unfortunate, I realize, hence the comment to call it out. 

Hoping that it doesn't spoil the NamedParam fun too much - was glad we didn't have more of these, do realize that if the super impl changed it could be a gotcha, but we would be making any such changes.

Was able to avoid another case of it below (and a type assertion) via check on `instanceof`

src/main/groovy/io/xh/hoist/cache/Cache.groovy
```diff
@@ -27,15 +29,27 @@ class Cache<K,V> extends BaseCache<V> {
...
+        @NamedParam Closure timestampFn = null,
+        @NamedParam boolean replicate = false,
+        @NamedParam boolean optimizeRemovals = false
+    ) {
+        super(svc, name, expireTime, expireFn, timestampFn, replicate, optimizeRemovals)
+
         if (replicate && !name) {
-            throw new RuntimeException("Cannot create a replicated cache without a unique name")
+            throw new IllegalArgumentException("Cannot create a replicated Cache without a unique name")
         }
```
amcclain, line 46: Unfortunately did not look like we could validate service-level uniqueness up front for `Cache` (I don't think)

amcclain: Also note that here I am deliberately _avoiding_ use of `this` - replication could be turned on at any time, should validate name even if not currently active.

src/main/groovy/io/xh/hoist/cache/CachedValue.groovy
```diff
@@ -9,23 +13,34 @@ import static java.lang.System.currentTimeMillis
...
+        // Intentional `this` below - super might override to false if !multiInstanceEnabled
+        _map = this.replicate ? svc.replicatedCachedValuesMap : svc.localCachedValuesMap
 
-    CachedValue(Map options) {
-       super(options)
-        if (!name) {
-            throw new RuntimeException("Cannot create a CachedValue without a unique name")
+        if (!name || _map.containsKey(name)) {
+            throw new IllegalArgumentException("Cannot create CachedValue ${name ? "with duplicate name '$name'" : "without a name"}")
         }
```
amcclain, line 40: This is a bit fussy I know, but still seemed worth it to actually check for uniqueness where we can

src/main/groovy/io/xh/hoist/cache/CachedValue.groovy
```diff
@@ -9,23 +13,34 @@ import static java.lang.System.currentTimeMillis
...
 /**
- * Similar to Cache, but a single value that can be read, written, and expired.
- *
+ * Similar to {@link Cache}, but a single value that can be read, written, and expired.
  * Like Cache, this object supports replication across the cluster.
  */
 class CachedValue<V> extends BaseCache<V> {
 
-    private final Map<String, Entry> _map
+    private final Map<String, Entry<V>> _map
```
amcclain, line 21: Adding `<V>` avoided a type assertion and some weak warnings in the code below

amcclain:
I am still hoping that we might spread the use of `@NamedVariant` around to other map-based ctors and APIs we have in Hoist. We have found it to be helpful in our large client project that uses it extensively.

I would like to look a bit more at the doc comment side of things - in the end would be most useful to doc the ctors vs. the properties.  Did not change that around now, however, as unclear if it could be done without duplication, and I was not in a hurry to add even more verbosity to this PR... 😉

amcclain:
Oh - important!  I did not actually RUN this code - need to head home and didn't want to leave it hanging - am happy to properly test in my local sandbox tomorrow to ensure I didn't bang anything up

src/main/groovy/io/xh/hoist/cache/CachedValue.groovy
lbwexler: I am not a huge fan of this one -- it actually doesn't give you any protection unless that CachedValue has already been set -- typically you will setup your CachedValues first before you add any content.

lbwexler:
I'll merge and tweak in my branch.  Great fixes

PR #391, opened 2024-09-04
Persistence Manager

Ryanseanlee:


ghsolomon (approved):

amcclain:
Closing this PR w/o merging as we don't need this for view manager (we are not using any JSON-based ALCs) and this change actually breaks our current implementation.  We rely on shared blobs to end up with an ACL of `*` and this changes that to `"*"` (quoted, as a JSON-serialized string value would be).

I'd say let's leave this along until we actually get to a point where JSONBlob is doing something with the ALC that's not just null vs *.

PR #392, opened 2024-09-05
Improvements to serialization efficiency of replicated `Cache` and `CachedValue`

lbwexler:
The hz event handler itself was triggering more serialization!  Skip it, unless we are going to do something with it.


TomTirapani:
Looks great! 👍

PR #398, opened 2024-09-11
Update `ClusterService` to use Hoist's `InstanceNotFoundException` class

ghsolomon:


ghsolomon:
@lbwexler looks like `InstanceNotFoundException` should have already been routine - was just an import issue in `ClusterService`

lbwexler:
OMG -- unbelieveable.  Need to figure out how to stop hiding my imports in intellij.  Import errors like this are killing us in both JS and Java.  

We should highlight this as something we need to be careful about as in a team and with CR

PR #399, opened 2024-09-11
Enhance Timer/Cache/CachedValues API

lbwexler:
+ New BaseService factory methods + Docs
+ Require/Check unique names for *all *service resources
+ Use NamedVariant for Timer
+ Improvements to `Timer` to avoid extra executions when primary instance changes

+ Unrelated cleanups to Timer declarations and deprecate `onTimer`
+ Logs should get archived on non-primary server as well

See also change on toolbox

### DIFF
--- a/grails-app/services/io/xh/hoist/config/ConfigService.groovy
+++ b/grails-app/services/io/xh/hoist/config/ConfigService.groovy
@@ -207,8 +207,7 @@ class ConfigService extends BaseService {
     }
 
     void fireConfigChanged(AppConfig obj) {
-        def topic = clusterService.getTopic('xhConfigChanged')
-        topic.publishAsync([key: obj.name, value: obj.externalValue()])
+        getTopic('xhConfigChanged').publishAsync([key: obj.name, value: obj.externalValue()])
     }
 
     //-------------------
--- a/src/main/groovy/io/xh/hoist/BaseService.groovy
+++ b/src/main/groovy/io/xh/hoist/BaseService.groovy
@@ -392,13 +392,24 @@ abstract class BaseService implements LogSupport, IdentitySupport, DisposableBea
         resourceNames << name
     }
 
+    /** @internal - for use by Cache */
+    Map getCacheReplicatedMap(String name) {
+        ClusterService.hzInstance.getReplicatedMap(hzName(name))
+    }
+
     /** @internal - for use by CachedValue */
     Map getReplicatedCachedValuesMap() {
-        _replicatedCachedValues ?= getReplicatedMap('cachedValues')
+        if (_replicatedCachedValues == null) {
+            _replicatedCachedValues = getReplicatedMap('cachedValues')
+        }
+        return _replicatedCachedValues
     }
 
     /** @internal - for use by CachedValue */
     Map getLocalCachedValuesMap() {
-        _localCachedValues ?= new ConcurrentHashMap()
+        if (_localCachedValues == null) {
+            _localCachedValues = new ConcurrentHashMap()
+        }
+        return _localCachedValues
     }
 }
--- a/src/main/groovy/io/xh/hoist/BaseService.groovy
+++ b/src/main/groovy/io/xh/hoist/BaseService.groovy
@@ -122,49 +122,48 @@ abstract class BaseService implements LogSupport, IdentitySupport, DisposableBea
     // Use static reference to ClusterService to allow access pre-init.
     //------------------------------------------------------------------
     /**
-     * Get a reference to a Hazelcast IMap.
+     * Create and return a reference to a Hazelcast IMap.
      *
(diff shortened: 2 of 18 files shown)


TomTirapani:
Looks good to me!

My main concern is that the timer having finished on primary does not _necessarily_ mean that the results of that timer have been replicated and that the time doesn't need to run on the new primary... thinking about very large datasets where it could potentially be brought down in the middle of serialisation. But that seems like an edge case that maybe we don't need to worry about

lbwexler:
Tom -- that's a really good point -- had not thought about that one

src/main/groovy/io/xh/hoist/util/Timer.groovy
```diff
@@ -109,6 +114,9 @@ class Timer {
     private java.util.Timer configTimer
 
 
+    private CachedValue<Date> _lastCompletedOnCluster
```
amcclain, line 117: Should we start using `Instant` as a general policy?

amcclain: @jskupsik has been our `Instant` advocate here :)

src/main/groovy/io/xh/hoist/util/Timer.groovy
```diff
@@ -287,15 +306,19 @@ class Timer {
...
-                if (wasForced) forceRun = false
-            }
+        if (!isRunning && (forceRun || isIntervalElapsed())) {
+            boolean wasForced = forceRun
+            doRun()
+            if (wasForced) forceRun = false
         }
     }
 
+    private boolean isIntervalElapsed() {
```
amcclain, line 316: Nit - `intervalHasElapsed`?

src/main/groovy/io/xh/hoist/util/Timer.groovy
lbwexler: yes -- probably just  being conservative with this change, but perhaps no time like the present

grails-app/services/io/xh/hoist/config/ConfigService.groovy
```diff
@@ -207,8 +207,7 @@ class ConfigService extends BaseService {
     }
 
     void fireConfigChanged(AppConfig obj) {
-        def topic = clusterService.getTopic('xhConfigChanged')
-        topic.publishAsync([key: obj.name, value: obj.externalValue()])
+        getTopic('xhConfigChanged').publishAsync([key: obj.name, value: obj.externalValue()])
```
amcclain, line 210: @lbwexler this is correct, right? No need to get topic from ClusterService, thinking this might have been an earlier implementation before we had on BaseService.  Just checking that I am not missing something special about this particular usage.

amcclain (commented):
Looks great, thanks!  Two follow-up questions - small one on if we should check useCluster and bigger one on naming of replicated obj getters (or rather, creators)

src/main/groovy/io/xh/hoist/cache/Cache.groovy
```diff
@@ -28,34 +26,35 @@ import static java.lang.System.currentTimeMillis
...
-        timer = new Timer(
-            owner: svc,
-            primaryOnly: replicate,
+        cullTimer = svc.createTimer(
+            name: "xh_${name}_cullEntries",
             runFn: this.&cullEntries,
             interval: 15 * MINUTES,
-            delay: true
+            delay: true,
+            primaryOnly: replicate
```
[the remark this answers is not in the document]
lbwexler: yep, good catch

src/main/groovy/io/xh/hoist/BaseService.groovy
```diff
@@ -109,37 +121,105 @@ abstract class BaseService implements LogSupport, IdentitySupport, DisposableBea
...
     //------------------------------------------------------------------
-    <K, V> IMap<K, V> getIMap(String id) {
-        ClusterService.hzInstance.getMap(hzName(id))
+    /**
+     * Get a reference to a Hazelcast IMap.
+     *
+     * @param name - unique name relative to all Caches, Timers and Hazelcast objects
+     *      associated with this service.
+     */
+    <K, V> IMap<K, V> getIMap(String name) {
```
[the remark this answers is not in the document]
lbwexler: no practical implications given this is all still so new-- just a matter of churn in our changelog

grails-app/services/io/xh/hoist/config/ConfigService.groovy
lbwexler: nice cleanup!

src/main/groovy/io/xh/hoist/util/Timer.groovy
lbwexler: gonna wait for a different PR to do Date to Instant cleanups.  Will keep it in mind.

PR #401, opened 2024-09-12
Enhance RestController to support multiple datasources

### LINKED ISSUE #400: RestController requires overrides to support multiple datasources
A client app is trying to map to a secondary datasource in a domain object, but defaults to the first datasource. 

Another client app works around this by overriding several RestController methods and wrapping them in `restTarget.withTransaction`, and references this Grails issue: https://github.com/grails/grails-core/issues/10383

We should be able to apply this pattern within RestController so that app's have this support out of the box.

saba-mo:
Resolves #400

### DIFF
--- a/grails-app/controllers/io/xh/hoist/RestController.groovy
+++ b/grails-app/controllers/io/xh/hoist/RestController.groovy
@@ -7,11 +7,9 @@
 
 package io.xh.hoist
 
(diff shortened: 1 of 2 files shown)


lbwexler:
Add Technical ChangeLog entry? Its a pretty core object

lbwexler (approved):

PR #402, opened 2024-09-13
Admin cluster tab refresh

lbwexler:


amcclain:
Did you mean for this to also include changes to alert banner? Looks like just timer changes + log exception change?

amcclain:
Sorry to be a pain, but if we could update PR title if we merge this - would be great as these PR titles end up in generated release content.  Thanks!

lbwexler:
Or course.

PR #422, opened 2024-11-22
Adjust App Log Name

lbwexler:
For easier identification/sorting of the main log with its siblings logs on a given date.


amcclain (approved):
All for alignment!

PR #425, opened 2024-12-02
Reconcile use of  handleException vs. logError for server side exception handling.

lbwexler:
We use both `handleException` and `logError` on the server.   But the former provides additional services (e.g. sanitization, recognition of "RoutineExceptions" and other conventions).  That logic should be used generally when logging our exceptions

Not sure how we want to do this syntactically -- integrate the use of ExceptionHandler into 'logSupport.logError'?   Favor the use of `xhExceptionHandler.handleException`?  Or write a new service method `BaseService.logException` that will call the latter in a more streamlined way?


lbwexler:
Going to merge this simple fix to close this ticket.

PR #426, opened 2024-12-12
Enhanced DistributedObjectAdmin to compare for inconsistencies between instances

jskupsik:


grails-app/init/io/xh/hoist/ClusterConfig.groovy
```diff
@@ -91,6 +91,8 @@ class ClusterConfig {
     Config createConfig() {
         def ret = new Config()
 
+        System.out.println("ClusterConfig [INFO] | ${multiInstanceEnabled ? 'Multi-instance is enabled - instances will attempt to cluster.' : 'Multi-instance is disabled - instances will avoid clustering.'}")
```
lbwexler, line 94: I think trying to mimic the log output is confusing --- can we just log this later in cluster service, when our logging is setup?

grails-app/services/io/xh/hoist/admin/DistributedObjectAdminService.groovy
```diff
@@ -0,0 +1,247 @@
...
+
+        return [*hzObjs, *resourceObjs].findAll{ it } as List<DistributedObjectInfo>
+    }
+    static class ListDistributedObjects extends ClusterRequest<List<DistributedObjectInfo>> {
+        List<DistributedObjectInfo> doCall() {
+            appContext.distributedObjectAdminService.listDistributedObjects()
+        }
+    }
+
+    void clearObjects(List<String> names) {
```
lbwexler, line 69: Lets call this clearHibernateCaches and be specific!   Could potentially combine with method below.

grails-app/services/io/xh/hoist/admin/DistributedObjectAdminService.groovy
```diff
@@ -0,0 +1,247 @@
...
+            adminStats = null,
+            error = null
+
+        try {
+            comparisonFields = obj.hasProperty('comparisonFields') ? obj.comparisonFields : null
+            adminStats = obj.hasProperty('adminStats') ? obj.adminStats : null
+        } catch (Exception e) {
+            def msg = 'Error extracting admin stats'
+            logError(msg, e)
+            error = "$msg | $e.message"
```
lbwexler, line 105: would prefer curly braces on $e.message.

grails-app/services/io/xh/hoist/admin/DistributedObjectAdminService.groovy
```diff
@@ -0,0 +1,247 @@
...
+        )
+    }
+
+    private List<DistributedObjectInfo> listDistributedObjects() {
+        // Services and their resources
+        Map<String, BaseService> svcs = grailsApplication.mainContext.getBeansOfType(BaseService.class, false, false)
+        def resourceObjs = svcs.collectMany { _, svc ->
+            [
+                // Services themselves
+                getInfo(obj: svc, name: svc.class.getName(), type: 'Service'),
```
lbwexler, line 47: think we could clarify with two methods that each just take a single object, and do type switching as needed -- e.g. getHoistInfo(Object obj) and getHzInfo(DistributedObject obj) -- symmetry!

grails-app/services/io/xh/hoist/cluster/ClusterService.groovy
```diff
@@ -185,8 +185,10 @@ class ClusterService extends BaseService implements ApplicationListener<Applicat
         def clazz
         try {
             clazz = Class.forName(Utils.appPackage + '.ClusterConfig')
+            System.out.println("ClusterService [INFO] | Found custom ClusterConfig at ${Utils.appPackage + '.ClusterConfig'}")
```
lbwexler, line 188: again, let's just print out the cluster config package path when logging is available?  We have a reference to an instance of the class and can easily print out its full canonical name.

grails-app/controllers/io/xh/hoist/admin/cluster/DistributedObjectAdminController.groovy
```diff
@@ -0,0 +1,32 @@
...
+@Access(['HOIST_ADMIN_READER'])
+class DistributedObjectAdminController extends BaseController {
+    def distributedObjectAdminService
+
+    def getDistributedObjectsReport() {
+        renderJSON(distributedObjectAdminService.getDistributedObjectsReport())
+    }
+
+    @Access(['HOIST_ADMIN'])
+    def clearObjects() {
```
lbwexler, line 21: see below re: naming of these two endpoints to be more specific.

src/main/groovy/io/xh/hoist/cache/Cache.groovy
```diff
@@ -221,15 +220,20 @@ class Cache<K, V> implements LogSupport {
...
+                name           : name,
+                type           : 'Cache' + (replicate ? ' (replicated)' : ''),
+                count          : size(),
+                latestTimestamp: _map.max { it.value.dateEntered }?.value?.dateEntered,
+                lastCullTime   : cullTimer.lastRunCompleted
         ]
     }
 
+    List getComparisonFields() {
+        if (!replicate) return null
```
lbwexler, line 233: super class returned empty list.  maybe better to use ternary and return super.getComparisonFields()

src/main/groovy/io/xh/hoist/cachedvalue/CachedValueEntry.groovy
```diff
@@ -30,6 +30,8 @@ class CachedValueEntry<T> implements KryoSerializable, LogSupport {
         this.uuid = UUID.randomUUID()
     }
 
+    static final UNINITIALIZED_CACHE_VALUE_ENTRY = new CachedValueEntry()
```
lbwexler, line 33: not sure why this change was made, but it worries me that if you ever tried to serialize one of these, I think it would fail with the null loggerName.  If we *do* need this, singleton, I might suggest  a different name, and also using the main constructor.

src/main/groovy/io/xh/hoist/cluster/DistributedObjectInfo.groovy
```diff
@@ -0,0 +1,49 @@
+package io.xh.hoist.cluster
+
+import io.xh.hoist.json.JSONFormat
+
+import static io.xh.hoist.util.Utils.getClusterService
+
+class DistributedObjectInfo implements JSONFormat {
+    // Absolute name of the object, make sure to use `svc.hzName(name)` on relative-named objects
```
lbwexler, line 8: I find these interleaved comments make it difficult to read the fairly straightforward property name

src/main/groovy/io/xh/hoist/cluster/DistributedObjectsReport.groovy
```diff
@@ -0,0 +1,48 @@
...
+
+    private Map<String, List<List<String>>> createBreaks() {
+        Map<String, List<List<String>>> breaks = [:]
+        info.groupBy { it.name }.each { name, infoObjs ->
+            [infoObjs, infoObjs].eachCombination { a, b ->
+                // Skip comparing objects to themselves.
+                if (a === b) return
+
+                if (!a.isMatching(b)) {
+                    if (!breaks[name]) breaks[name] = []
```
lbwexler, line 41: withDefault on line 34 makes this a bit cleaner

src/main/groovy/io/xh/hoist/cachedvalue/CachedValueEntry.groovy
jskupsik: When we create a CachedValue, we initialize its `entry` value with a null CachedValueEntry that is never replicated. 

```groovy
    private CachedValueEntry<V> entry = new CachedValueEntry<V>(null, loggerName)
```
[link to excerpt above](https://github.com/xh/hoist-core/blob/develop/src/main/groovy/io/xh/hoist/cachedvalue/CachedValue.groovy#L68)

This results in different `dateEntered` values between instances. It hasn't mattered before, but this is a value that is otherwise consistent and so I want to compare it between instances. The point of UNINITIALIZED_CACHE_VALUE_ENTRY was to have the `dateEntered` be null in this case, so the values will not be out of sync.

jskupsik: Also - just realized this - a bigger sample of the current code above:
```groovy
    private final String loggerName
    private final ITopic<CachedValueEntry<V>> topic
    private CachedValueEntry<V> entry = new CachedValueEntry<V>(null, loggerName)
    
    /** @internal - do not construct directly - use {@link BaseService#createCachedValue}. */
    @NamedVariant
    CachedValue(
        @NamedParam(required = true) String name,
        @NamedParam(required = true) BaseService svc,
        @NamedParam Object expireTime = null,
        @NamedParam Closure expireFn = null,
        @NamedParam Closure timestampFn = null,
        @NamedParam Boolean replicate = false,
        @NamedParam Closure onChange = null
    ) {

        this.name = name
        this.svc = svc
        this.expireTime = expireTime
        this.expireFn = expireFn
        this.timestampFn = timestampFn
        this.replicate = replicate

        // Allow fine grain logging for this within namespace of owning service
        loggerName = "${svc.instanceLog.name}.CachedValue[$name]"
```

Isn't `loggerName` already being initialized to null in the current code, since the inline field initialization like declaring `entry` runs before the constructor that sets `loggerName`?

grails-app/services/io/xh/hoist/admin/DistributedObjectAdminService.groovy
jskupsik: The issue I had with this was the fact that resource objects don't always know their full name - they often use a relative name. So a fully-symmetrical `getHoistInfo(Object obj)` might not have the info necessary to create a resource object entry. However, I can make it more symmetrical by taking object (required parameter) outside of the args (optional params).

grails-app/services/io/xh/hoist/admin/DistributedObjectAdminService.groovy
```diff
@@ -0,0 +1,264 @@
...
+            ]
+        },
+        // Distributed objects
+            hzObjs = clusterService
+                    .hzInstance
+                    .distributedObjects
+                    .findAll { !(it instanceof ExecutorServiceProxy) }
+                    .collect { getHzInfo(it) }
+
+        return [*hzObjs, *resourceObjs].findAll{ it } as List<DistributedObjectInfo>
```
lbwexler, line 61: nit: can be findAll()

grails-app/services/io/xh/hoist/admin/DistributedObjectAdminService.groovy
```diff
@@ -0,0 +1,264 @@
...
+            names = hibernateCaches.collect { it.getName() }
+        }
+
+        names.each { name ->
+            def obj = hibernateCaches.find { it.getName() == name }
+            if (obj) {
+                obj.clear()
+                logInfo("Cleared " + name)
+            } else {
+                logWarn('Cannot clear object - unsupported type', name)
```
lbwexler, line 82: obsolete log text.

PR #427, opened 2024-12-12
Separate configs for support and feedback emails

PeteDarinzo:
Currently, the email address configured for support is used for feedback notification emails. Applications may want to add additional recipients to receive application feedback emails, but not support requests.


lbwexler:
Thinking about this -- should we have a fallback to the support email?

lbwexler:
Closing this -- we came up with a different solution at the client that was requesting this.  Not clear how generally useful this new feature would be.

PR #428, opened 2024-12-19
New ViewService  + Fix bug with memory monitoring

lbwexler:


grails-app/controllers/io/xh/hoist/impl/XhController.groovy
```diff
@@ -176,6 +176,11 @@ class XhController extends BaseController {
         renderJSON(ret.formatForClient())
     }
 
+    def findJsonBlob(String type, String name) {
+        def ret = jsonBlobService.find(type, name)
+        renderJSON(ret?.formatForClient())
```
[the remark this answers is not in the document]
ghsolomon: Also realized this only returns a blob if it is **owned** by the user, whereas the `listJsonBlobs` endpoints returns blobs for which the user `passesAcl`.  I see why you did this, but I wonder if it could be confusing-- i.e. "Why do I see a blob in the list of all blobs, but when I try to look it up directly using the `findJsonBlob` endpoint, nothing is returned?"

grails-app/services/io/xh/hoist/jsonblob/JsonBlobService.groovy
```diff
@@ -48,15 +53,15 @@ class JsonBlobService extends BaseService implements DataBinder {
...
-            if (data.containsKey('meta')) data.meta = serialize(data.meta)
+        return updateInternal(blob, data, username)
+    }
 
-            bindData(blob, data)
-            blob.save()
-        }
-        return blob
+    @Transactional
+    JsonBlob createOrUpdate(String type, String name, Map data, String username = username) {
```
ghsolomon, line 60: Does this work for blobs that are not owned by the authenticated user?  In the context of views, blobs describing "Global" views with `null` owners?

ghsolomon: I think I just misunderstood it - looks good to me :)

lbwexler:
cr: greg

PR #437, opened 2025-01-15
Fix `JsonBlob` unique name validation

ghsolomon:


grails-app/domain/io/xh/hoist/jsonblob/JsonBlob.groovy
```diff
@@ -41,7 +41,17 @@ class JsonBlob implements JSONFormat {
...
         acl nullable: true
-        name maxSize: 255, blank: false, unique: ['owner', 'type', 'archivedDate']
+        name maxSize: 255, blank: false, validator: { val, obj ->
+            if (where {
+                name == val &&
+                    type == obj.type &&
+                    archivedDate == obj.archivedDate &&
+                    owner == obj.owner &&
+                    token != obj.token
+            }.count() > 0) {
```
lbwexler, line 51: for what its worth, I think you could use the exists() method on the DetachedCriteria returned here.....

lbwexler: also, would be useful to have a small comment noting the unique constraint was not working

grails-app/domain/io/xh/hoist/jsonblob/JsonBlob.groovy
```diff
@@ -42,14 +42,25 @@ class JsonBlob implements JSONFormat {
...
-                name == val &&
-                    type == obj.type &&
-                    archivedDate == obj.archivedDate &&
-                    owner == obj.owner &&
-                    token != obj.token
-            }.count() > 0) {
-                return 'default.not.unique.message'
+            // Unique constraint validation across nullable fields can be unreliable.
+            // See https://github.com/grails/grails-data-mapping/issues/1585
+
```
lbwexler, line 47: Greg and I are going to discuss this -- want to understand what we are getting from the separate clauses.

Also, will just argue that this should be in a private static method, so it does not overwhelm the entire constraints block

PR #438, opened 2025-01-21
Support Severity levels for Track Logging.

lbwexler:


lbwexler:
CR: John Haynes

grails-app/init/io/xh/hoist/BootStrap.groovy
```diff
@@ -72,7 +72,8 @@ class BootStrap implements LogSupport {
                     enabled: true,
                     logData: false,
                     maxDataLength: 2000,
-                    maxRows: [default: 10000, limit: 25000, options: [1000, 5000, 10000, 25000]]
+                    maxRows: [default: 10000, limit: 25000, options: [1000, 5000, 10000, 25000]],
+                    levels: [[user: '*', category: '*', severity: 'INFO']]
```
amcclain, line 76: We hit a bug with @RyanMG on his newly bootstrapped Toolbox instance - looks like this bootstrapped config is incorrect, should be `username`, fails badly breaking tracking.  We'll look to fix this now.

PR #439, opened 2025-01-24
Add `JsonSearchService` and controller

cnrudd:
Pair with Hoist React PR: https://github.com/xh/hoist-react/pull/3911

Docs for json-path and json-patch libs used:
-------------------

https://github.com/java-json-tools/json-patch
https://github.com/json-path/JsonPath

JSON Patch Standards Docs: https://jsonpatch.com/
JSON Path Standards Docs: https://goessner.net/articles/JsonPath/
JSON Patch RFC: https://datatracker.ietf.org/doc/html/rfc6902


grails-app/services/io/xh/hoist/admin/JsonSearchService.groovy
```diff
@@ -0,0 +1,107 @@
...
+            Option.ALWAYS_RETURN_LIST
+        ).build()
+
+    private Configuration nodeSearchValuesConf = Configuration.builder()
+        .options(
+            Option.ALWAYS_RETURN_LIST
+        ).build()
+
+    List<Map> searchBlobs(String path) {
+        List<JsonBlob> matches = JsonBlob.list().findAll { hasPathMatch(it.value, path) }
```
lbwexler, line 49: I think we should skip archived blobs.

grails-app/services/io/xh/hoist/admin/JsonSearchService.groovy
```diff
@@ -0,0 +1,107 @@
...
+            Option.AS_PATH_LIST,
+            Option.ALWAYS_RETURN_LIST
+        ).build()
+
+    private Configuration nodeSearchValuesConf = Configuration.builder()
+        .options(
+            Option.ALWAYS_RETURN_LIST
+        ).build()
+
+    List<Map> searchBlobs(String path) {
```
lbwexler, line 48: Am very confused as to why these don't need to be marked @ReadOnly

amcclain: They're currently only being called in the context of a controller call, yeah?

grails-app/services/io/xh/hoist/admin/JsonSearchService.groovy
amcclain: I was also wondering if we needed to support a `type` query param and pull that through the UI in a general way - not sure it's really necessary, but the unqualified `.list()` did make me think of it

lbwexler: was thinking the same

amcclain: I'd be OK with waiting to see if we need to figure out a type param (or support for additional params in general).  Right now the main JSONBlob admin page loads all blobs, so listing them here had better not cause any major issues.

But I do think skipping archive is a better default - I can mix that in now.  If we find we need ability to search archived blobs, we can revisit the default or figure out a treatment for params as per above.

amcclain (approved):
As per matching HR review, looks great, eager to get it in.

PR #440, opened 2025-01-28
Improve shutdown

lbwexler:


grails-app/controllers/io/xh/hoist/impl/XhController.groovy
```diff
@@ -297,12 +297,7 @@ class XhController extends BaseController {
...
             appCode: Utils.appCode,
             appVersion: Utils.appVersion,
-            appBuild: Utils.appBuild,
-            // TODO - this is a temporary measure to ensure that clients running hoist-react < 67
-            //      prompt for upgrade when a new app release is deployed with hoist-core >= 21.
-            //      Going forward clients will read these instructions from `xh/environmentPoll`.
-            //      Once a majority of key apps have been upgraded, we should remove this key.
-            mode: configService.getMap('xhEnvPollConfig').onVersionChange
+            appBuild: Utils.appBuild
         )
```
lbwexler, line 301: this is the one unrelated change here, but would like to include this in this major HoistCore version

grails-app/services/io/xh/hoist/cluster/ClusterService.groovy
```diff
@@ -62,8 +68,27 @@ class ClusterService extends BaseService implements ApplicationListener<Applicat
...
+        ] as LifecycleListener)
+    }
+
+    /**
+     * Called by Framework to shutdown the underlying Hazelcast instance.
+     * @internal
+     */
+    static void shutdownHazelcast() {
+        shutdownInProgress = true
+        hzInstance.shutdown()
```
amcclain, line 91: I'd be happy to have a log statement here and in the matching Timer code - just to confirm that these get run and have a little bit of visibility into it, help us remember these are here.

src/main/groovy/io/xh/hoist/util/Timer.groovy
```diff
@@ -48,6 +48,7 @@ import static java.lang.System.currentTimeMillis
 class Timer implements LogSupport {
 
     private static Long CONFIG_INTERVAL = 15 * SECONDS
+    private static boolean isShutdown = false
```
amcclain, line 51: Minor suggestion - rename to `shutdownInProgress` for consistency with ClusterService

grails-app/services/io/xh/hoist/cluster/ClusterService.groovy
```diff
@@ -62,8 +68,27 @@ class ClusterService extends BaseService implements ApplicationListener<Applicat
      * Called by Framework to initialize the Hazelcast instance.
      * @internal
      */
-    static initializeInstance() {
+    static void initializeHazelcast() {
         hzInstance = Hazelcast.newHazelcastInstance(clusterConfig.createConfig())
+        getHzConfig().addListenerConfig(new ListenerConfig())
+        hzInstance.lifecycleService.addLifecycleListener([
+            stateChanged: { LifecycleEvent e ->
+                // If hz shutdown *not* initiated by app, need to propagate to app/JVM
```
amcclain, line 76: Maybe clarify when this could happen / has been observed to happen?  We were seeing specific cases around OOM errors that affected HZ only, is that right?  I can imagine wondering in a year or two...

amcclain (commented):
Looks great, would be happy to get this merged, offering some very minor suggestions

src/main/groovy/io/xh/hoist/util/Timer.groovy
lbwexler: 👍🏻

grails-app/services/io/xh/hoist/cluster/ClusterService.groovy
lbwexler: 👍🏻

grails-app/services/io/xh/hoist/cluster/ClusterService.groovy
lbwexler: 👍🏻

PR #441, opened 2025-02-03
Remove default use of LDAP_MATCHING_RULE_IN_CHAIN

ghsolomon:
Client reported that our use of `LDAP_MATCHING_RULE_IN_CHAIN` has a negative performance impact on their LDAP databases and is unnecessary when querying groups without deep nesting.  Additionally, it is proprietary to MS and therefore not suitable for other LDAP implementations. This change recurses LDAP directories with its own code by default but provides a flag on `xhLdapConfig` to enable the previous implementation in cases where an AD instance uses deeply nested groups.

### DIFF
--- a/grails-app/services/io/xh/hoist/ldap/LdapService.groovy
+++ b/grails-app/services/io/xh/hoist/ldap/LdapService.groovy
@@ -159,7 +159,7 @@ class LdapService extends BaseService {
         config.useMatchingRuleInChain ?
             // See class-level comment regarding this AD-specific query
             searchMany("(|(memberOf=$dn) (memberOf:1.2.840.113556.1.4.1941:=$dn))", LdapPerson, strictMode) :
-            lookupGroupMembersRecursive(dn, strictMode)
+            lookupGroupMembersRecursive(dn, strictMode).unique { it.distinguishedname }
     }
 
     private List<LdapPerson> lookupGroupMembersRecursive(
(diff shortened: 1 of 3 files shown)


grails-app/services/io/xh/hoist/ldap/LdapService.groovy
```diff
@@ -156,8 +156,24 @@ class LdapService extends BaseService {
...
+
+    private List<LdapPerson> lookupGroupMembersRecursive(
+        String dn,
+        boolean strictMode,
+        Set<String> visitedGroups = new HashSet<String>()
+    ) {
+        if (!visitedGroups.add(dn)) return []
+        List<LdapPerson> users = searchMany("(memberOf=$dn)", LdapPerson, strictMode)
+        List<LdapGroup> groups = searchMany("(memberOf=$dn)", LdapGroup, strictMode)
+        groups.each { LdapGroup group ->
```
lbwexler, line 173: Think this looks generally fine, but a bit worried about the efficiency if we were handling huge user lists (I know its unlikely).  Looks like groovy's unique is n^2! 
(https://softwareengineering.stackexchange.com/questions/202113/whats-the-computational-complexity-of-the-groovy-unique-method)

Probably safest to pass a collector Set in to recursive method that gathers the unique users, and just convert it to a list in the caller.   No need for the recursive method to return anything

cnrudd: Looks like the n^2 is still true on the version we are using:
https://github.com/apache/groovy/blob/GROOVY_3_0_23/src/main/java/org/codehaus/groovy/runtime/DefaultGroovyMethods.java#L1324

vs

version from 2013 Stack Exchange post: https://github.com/groovy/groovy-core/blob/6232fdb2872c13f94d5e9fd24ddf2d16fd61ee2d/src/main/org/codehaus/groovy/runtime/DefaultGroovyMethods.java#L1031

I was hoping it might have changed in 12 years since, but it has not.

PR #442, opened 2025-02-05
Add logging wrappers to LdapService calls

amcclain:


grails-app/services/io/xh/hoist/ldap/LdapService.groovy
```diff
@@ -192,7 +207,7 @@ class LdapService extends BaseService {
         List<T> ret = cache.get(key)
         if (ret != null) return ret
 
-        withDebug(["Querying LDAP", [host: host, filter: filter]]) {
+        withTrace(["Querying LDAP", [host: host, filter: filter]]) {
```
amcclain, line 210: This is the lowest-level query call which will be executed (potentially a number of times) as part of all the higher-level requests. Thinking this makes the most sense to be on trace.

amcclain:
Opened as a draft b/c I have not yet run this code, but I would like to get this in with the next release if we can verify.  I think the changes should be clear and I did my best to be careful, but always want to be up-front about code that's not yet been run!

(Discussed need for proper Hoist-side LDAP test env on Slack yesterday - can aim to get that done, or fold this into on-prem testing.)

lbwexler (approved):
Looks great -- nice pickup.

PR #446, opened 2025-03-03
Make Exception Message Parsing in JSONClient more flexible

lbwexler:


src/main/groovy/io/xh/hoist/http/JSONClient.groovy
```diff
@@ -146,32 +146,44 @@ class JSONClient {
...
 
-            // Message is required
-            if (msg instanceof String) {
+            // [1] We have a valid json object (preferred)
+            Map obj = safeParseObject(content)
+            if (obj) {
+                String msg = obj.msg instanceof String ? obj.msg : 'Unknown Error'
 
                 // Try to rehydrate exception of certain known and present classes
+                def className = obj.class
```
febbraiod, line 160: className is ex?.className previously, but now it's obj.class?  Intentional?

lbwexler:
great catch Don!  Need to test this a bit more.

PR #448, opened 2025-03-11
Convert `BaseAuthenticationService.whitelistURIs` to protected getter

amcclain:
- Support targeted implementation overrides of just the URIs themselves, without having to override `isWhitelist()`.

### DIFF
--- a/src/main/groovy/io/xh/hoist/security/BaseAuthenticationService.groovy
+++ b/src/main/groovy/io/xh/hoist/security/BaseAuthenticationService.groovy
@@ -159,12 +159,14 @@ abstract class BaseAuthenticationService extends BaseService {
      * that URI as otherwise SSO-based apps will not have a first shot at installing a user on the
      * session within their completeAuthentication() implementations.
      */
-    protected List<String> whitelistURIs = [
-        '/ping',  // legacy alias for /xh/ping (via UrlMappings)
-        '/xh/login',
-        '/xh/logout',
-        '/xh/ping',
-        '/xh/version',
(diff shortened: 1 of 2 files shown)


src/main/groovy/io/xh/hoist/security/BaseAuthenticationService.groovy
```diff
@@ -159,12 +159,14 @@ abstract class BaseAuthenticationService extends BaseService {
      * that URI as otherwise SSO-based apps will not have a first shot at installing a user on the
      * session within their completeAuthentication() implementations.
      */
```
lbwexler, line 161: I feel like we have client code *somewhere* that is actually just mutating this array in a way that would not be possible after this change.

lbwexler: So I found it -- the code is below, redacted.  So it could be rewritten as an override.

`
AuthenticationService() {
super()
whitelistURIs.push('/xxxUser/photos')
}
`

amcclain: Honestly it hadn't occurred to me, just mutating the list. Does that seem like....a good pattern, better than override? (honest question)

lbwexler:
See you closed this, which I agree with, but just to answer the question
--- maybe not the best pattern, but it does get us a cached, extensible
list -- in this case seems fairly functional for what we are trying to do,
if the list can be computed statically when object is constructed.

PR #450, opened 2025-03-15
Added new default `xhClusterObjectBreaksMonitor`

amcclain:


amcclain:
I threw this together quick , but warrants a closer look to ensure I have the impl correct and to review any concern over this "flapping" or creating more noise than we expect.  We know there can be transient breaks - we expect them to be very short-lived - this monitor might happen to catch those, but our general debouncing on alerting should handle that.

Motivation is - were we to have some long-running break across instances - we'd currently need to look at admin console to know.

lbwexler:
Looks great.  My only super nitty thing is I might call it just `xhClusterBreaksMonitor` and I might default it to warning

lbwexler (approved):
Made the slight rename, but other than that great

PR #451, opened 2025-03-18
Log client app version in websocket to be displayed for admins

RyanMG:
Ticket https://github.com/xh/hoist-react/issues/3947

Server side update for https://github.com/xh/hoist-react/pull/3948

### DIFF
--- a/grails-app/services/io/xh/hoist/websocket/WebSocketService.groovy
+++ b/grails-app/services/io/xh/hoist/websocket/WebSocketService.groovy
@@ -137,9 +137,6 @@ class WebSocketService extends BaseService implements EventPublisher {
         def msgJSON = deserialize(message)
 
         if (msgJSON.topic == HEARTBEAT_TOPIC) {
-            def msgClientVersion = msgJSON.clientAppVersion;
-            channel.setClientAppVersion(msgClientVersion as String)
-
             sendMessage(channel, HEARTBEAT_TOPIC, 'pong')
         } else {
             notify(MSG_RECEIVED_EVENT, [channel: channel, topic: msgJSON.topic, data: msgJSON.data])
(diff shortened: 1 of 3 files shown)


grails-app/services/io/xh/hoist/websocket/WebSocketService.groovy
```diff
@@ -137,6 +137,9 @@ class WebSocketService extends BaseService implements EventPublisher {
         def msgJSON = deserialize(message)
 
         if (msgJSON.topic == HEARTBEAT_TOPIC) {
+            def msgClientVersion = msgJSON.clientAppVersion;
+            channel.setClientAppVersion(msgClientVersion as String)
```
RyanMG, line 141: This coercion seems to work. I tried an update to now send any value from the client, and this did not crash or fail. But please let me know if there are better ways to access / set this data.

RyanMG:
@lbwexler updated.

1. Extract client app version off handshake rather than heartbeat.

src/main/groovy/io/xh/hoist/websocket/HoistWebSocketChannel.groovy
```diff
@@ -90,6 +94,23 @@ class HoistWebSocketChannel implements JSONFormat, LogSupport {
         return (String) session.attributes[IdentityService.APPARENT_USER_KEY] ?: 'unknownUser'
     }
 
+    private static Map<String, String> getSocketConnectionQueryParams(String query) {
```
RyanMG, line 97: This is an ideal method to add testing to. Is there testing established for hoist-core I can add to? I am not seeing it, but I am not as familiar with grails as Spring Boot.

lbwexler: nothing here at this time -- thanks for raising.

lbwexler (approved):

PR #452, opened 2025-03-20
Improvement to rendering of NO_CONTENT successful results

lbwexler:


src/main/groovy/io/xh/hoist/cluster/ClusterTaskException.groovy
```diff
@@ -0,0 +1,24 @@
...
+    String causeClassName
+    String causeMessage
+    String causeAsJson
+    Integer causeStatusCode
+
+    ClusterTaskException(Throwable cause) {
+        super()
+        causeClassName = cause.class.name
+        causeMessage = cause.message
+        causeAsJson = JSONSerializer.serialize(cause)
```
amcclain, line 21: Would there be any value in doing this serialization safely, so some unexpected/strange exception that could not be JSON-serialized didn't itself throw an exception here?

amcclain (approved):
Great! Nice that you could simplify it back a bit to the single object, and that's clever to use `null` vs `"null"` and fit it into the overall "if it's a string will be assumed to be JSON" strategy.

Asked the one question about cause exception serialization in the new `ClusterTaskException` wrapper, but either way am good to merge.

src/main/groovy/io/xh/hoist/cluster/ClusterTaskException.groovy
lbwexler: Merging, but will think on this additional change over the weekend!  Thanks for the careful look!

PR #467, opened 2025-05-08
Integrate Client Errors and Feedback with Tracking

lbwexler:
See PRs in Hoist React and Hoist Core

Big remaining item is throttling -- restore for 'Client Error' tracking? Extend to all track messages?


amcclain (approved):
Huge cleanup!

amcclain:
This worked locally and covered the single-arg variation, but did not push as was not sure how you feel about it

```

    /**
     * Create a new track log entry.
     * User, browser, environment, and datetime will be set automatically if not provided.
     */
    @NamedVariant
    void track(
        /** A concise description of whatever action being tracked. */
        @NamedParam(required = true) String msg,
        /** Category for organizing and querying over related actions. */
        @NamedParam String category = 'Default',
        /** Additional data payload to store with the track log (will be serialized as JSON). */
        @NamedParam Object data = null,
        /**
         * A list of keys from the data object that should have their values included in log file
         * written for this entry, or `true` to log all key/values.
         */
        @NamedParam Object logData = null,
        /**
         * The username of the user associated with this action. Defaults to currently authenticated
         * user, but can be specified to capture user when called from an async process, outside the
         * context of a request.
         */
        @NamedParam String username = null,
        /**
         * If impersonating, the username of the user being impersonated. Set automatically when
         * applicable, provided here for async usage outside of a request (as with `username`).
         */
        @NamedParam String impersonating = null,
        /**
         * The severity of this message. Defaults to TrackSeverity.INFO, but can be set to any
         * value in the TrackSeverity enum.
         */
        @NamedParam Object severity = TrackSeverity.INFO,
        @NamedParam String correlationId = null,
        @NamedParam String url = null,
        /** Time time associated with the start of the action. Defaults to current time. */
        @NamedParam Long timestamp = null,
        /** Duration of the tracked action in milliseconds, if applicable. */
        @NamedParam Integer elapsed = null,
        /** Unique client-side tabId, maintained by hoist-react for the life of a browser tab. */
        @NamedParam String tabId = null,
        /** Unique client-side loadId, maintained by hoist-react for each load of the client app. */
        @NamedParam String loadId = null
    ) {
        trackAll([
            [
                msg          : msg,
                category     : category,
                correlationId: correlationId,
                data         : data,
                logData      : logData,
                username     : username,
                impersonating: impersonating,
                severity     : severity,
                url          : url,
                timestamp    : timestamp,
                elapsed      : elapsed,
                tabId        : tabId,
                loadId       : loadId
            ]
        ] as Collection<Map>)
    }
```

CHANGELOG.md
```diff
@@ -2,6 +2,22 @@
...
+### 💥 Breaking Changes (upgrade difficulty: 🟢 LOW - may require event handling adjustment)
+
+* The `xhFeedbackReceived` and `xhClientErrorReceived` events have been removed. Clients should
+  use the `xhTrackReceived` event instead, and filter messages based on the `TrackLog.category`
+  property.
+
+### ⚙️ Technical
+
+* The `xh_feedback` and `xh_client_error` tables are obsolete and may be archived or removed from
+  the application database.
```
amcclain, line 19: Should we state it a bit more strongly that any existing entries across both tables would be orphaned there?

lbwexler: Will do -- thought about that.

lbwexler: Regarding the named variant -- the only problem is that the multiple arg case is the one that is the main entry point for the main usage -- the client calls.  The defaults shown here will not be applied to that call, but a reader of the code might reasonably assume they would be.  Or course that really is an internal call for the client only, and could also be marked as @internal, or even renamed.

I think if we really want to we could also have the main methods take a 'TrackRequest' query object.  For backward compatibility

lbwexler: I think on balance, probably worth going with your change

PR #469, opened 2025-05-14
Expose TrackSeverity in tracking log statements + log at that severity.

lbwexler:
Plus some related cleanups in the logging.

### DIFF
--- a/grails-app/services/io/xh/hoist/track/TrackLoggingService.groovy
+++ b/grails-app/services/io/xh/hoist/track/TrackLoggingService.groovy
@@ -43,11 +43,9 @@ class TrackLoggingService extends BaseService {
     }
 
     void logEntry(TimestampedLogEntry entry) {
-        // Write directly to main log,  Get as close as possible to timestamp and make actual time of timestamp available
-        writeLog(this, entry.severity, [
-            _timestamp: entry.timestamp.format('HH:mm:ss.SSS'),
-            *:entry.message
-        ])
+        // Write directly to main log. Get as close as possible to sort order of actual
+        // Actual time is also included in the message
+        writeLog(this, entry.severity, entry.message)
         queue.add(entry)
         drainQueue()
     }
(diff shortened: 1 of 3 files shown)


grails-app/services/io/xh/hoist/track/TrackLoggingService.groovy
```diff
@@ -42,7 +43,11 @@ class TrackLoggingService extends BaseService {
     }
 
     void logEntry(TimestampedLogEntry entry) {
-        logInfo(entry.message)
+        // Write directly to main log,  Get as close as possible to timestamp and make actual time of timestamp available
+        writeLog(this, entry.severity, [
+            _timestamp: entry.timestamp.format('HH:mm:ss.SSS'),
+            *:entry.message
+        ])
```
amcclain, line 50: I don’t remember (and realize this is not new), but why do we go to the trouble of the queue-based debouncing for the dedicated log but write directly into the main log?

grails-app/services/io/xh/hoist/track/TrackLoggingService.groovy
```diff
@@ -54,13 +59,37 @@ class TrackLoggingService extends BaseService {
         for (def e = queue.peek(); e && intervalElapsed(DEBOUNCE_INTERVAL, e.timestamp); e = queue.peek()) {
             def entry = queue.remove()
             try {
-                orderedLog.logInfo(entry.message)
+                // Write directly to dedicated log  Add timestamp, severity as this log has a very minimal layout
+                writeLog(orderedLog, entry.severity, [
+                    _timestamp: entry.timestamp.format('yyyy-MM-dd HH:mm:ss.SSS'),
+                    _severity: entry.severity,
```
amcclain, line 65: Why do we add the `_severity` here and not in the main log?  

Given that the logger is selected by severity in `writeLog` below, what's the extra advantage of this?

grails-app/services/io/xh/hoist/track/TrackLoggingService.groovy
lbwexler: the main log is always getting entries from other sources -- so no way for us to get these entries into their sorted order with the debounce.  so deemed best to submit right away.

amcclain: Ah right - this logger debouncing isn't a performance/load thing - like it is for superficially similar debounced writebacks to the database.  It's about minimizing out-of-time-order track statements in the tracking log itself.  I totally forgot that - re-reading the class level doc comment I see we document exactly that.  Sorry for the noise.

amcclain (approved):

PR #475, opened 2025-08-19
Default newly promoted global views to "pinned by default"

amcclain:
* Updated ViewManager to set newly promoted global views to be "pinned by default" for all users,
  avoiding an extra step for the common case where a global view should be readily visible to all.

### DIFF
--- a/grails-app/services/io/xh/hoist/view/ViewService.groovy
+++ b/grails-app/services/io/xh/hoist/view/ViewService.groovy
@@ -156,8 +156,14 @@ class ViewService extends BaseService {
     /** Make a view global */
     Map makeGlobal(String token, String username = username) {
         def existing = jsonBlobService.get(token, username),
-            meta = parseObject(existing.meta)?.findAll { it.key != 'isShared' },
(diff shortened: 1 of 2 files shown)


grails-app/services/io/xh/hoist/view/ViewService.groovy
```diff
@@ -156,8 +156,14 @@ class ViewService extends BaseService {
     /** Make a view global */
     Map makeGlobal(String token, String username = username) {
         def existing = jsonBlobService.get(token, username),
-            meta = parseObject(existing.meta)?.findAll { it.key != 'isShared' },
-            ret = jsonBlobService.update(token, [owner: null, acl: '*', meta: meta], username)
+            meta = parseObject(existing.meta)?.findAll { it.key != 'isShared' } ?: [:]
+
+        // When promoting to a global view, pin to all user menus by default, w/expectation that
```
lbwexler, line 161: I think we should make this an optional checkbox in the UI -- can default to true.  I think that will give more flexibility, but also force the user to think about it and understand it.

PR #477, opened 2025-08-21
Use User Agent 'Hint' Headers in Browser detection.

lbwexler:
Was motivated by client using Island browser, which is now only publishing its identity using the new headers.
Also some issues with device detection for this browser, which also requires the new headers.

This also unwinds some very ancient iOS homescreen code that we believe to be obsolete.

Note that we have probably reached our limits with our custom UA sniffing -- but could not see an easily dropped in java library -- `ua-parser` and `Yauaa` would require some work around multi-threading.  For now -- just extending and simplifying our existing approach seemed pragmatic.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,13 +1,20 @@
 # Changelog
 
-## 31.2.0 - 2025-08-27
+## 32.0.0 - 2025-08-27
+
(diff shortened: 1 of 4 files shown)


lbwexler:
cr: colin.  Merging, will do some additional testing of ios homescreen on deployed version.

PR #483, opened 2025-09-30
Disable MemoryMonitoringService support for past instances on localhost

amcclain:
- I was confused looking at the past instances drop-down on a non-prod-but-deployed application instance at a client, where we were troubleshooting some unexpected restarts due to OOM issues. I was seeing more past instance records available than I expected, and with weird timestamps. Realized eventually that these were from local development efforts running in that env and persisting to the same DB as the deployed app.
- Do not think we need/want this support in local dev mode.


lbwexler:
good changes.

PR #486, opened 2025-10-24
CollectionUtils + Enhancements to EmailService

lbwexler:
pending testing of email service


TomTirapani:
Looks great Lee! Great to have more performant versions of `collect` and `collectEntries` in here too 👍

PR #492, opened 2025-12-22
Enhance Client Activity and Web Sockets to be Client App Code aware

lbwexler:


jskupsik (approved):

CHANGELOG.md
```diff
@@ -7,6 +7,9 @@
 * The generic signature of `CacheEntry` has changed to support a generic key type, rather than just
  string.  Apps that are declaring this type will need to be adjusted to make the key type explicit
 
+### 🎁 New Features
+
+* Support for tracking `clientAppCode` in admin activity tracking.
```
amcclain: We typically provide sample SQL statements to add the new column, as not all envs are setup to do so automatically - see notes on v30.0.0 for an example.  

We haven't been doing so for indices - not sure if we should be equally specific there, or just let them be auto-managed when they can and risk doing without when they can't. Believe that very thing has bit us though in specific client envs - @lbwexler do you think it's worth including as well?

lbwexler: mentioned both in changelog, great catch

PR #496, opened 2025-12-31
Updated Misc server dependencies

lbwexler:


lbwexler:
Colin will test this upgraded poi library pre-release in a client app that is using it extensively

lbwexler:
cr: crudd (pending)

PR #498, opened 2025-12-31
New Decorators @RequiresAny and @RequiresEvery

lbwexler:


lbwexler:
Final idea is just to prefix these with `Access` -- bit longer, but nice to have the entire family with the same prefix.

PR #501, opened 2026-01-16
Enhance WebSocketService for cluster

lbwexler:
@jskupsik -- let me know your thoughts?  I realize this framing goes a little against what we originally agreed upon, but I think its much clearer.  More of a breaking change, but worth it.

At the current moment, this is really only a breaking change for two multi-instance apps that I know of.

Regarding the closure form of selecting channel.  I think we could still add something like that, but it is going to have be done differently with an explicit class name, and some parameters for it, I think.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,9 @@
    will be removed in a future version of Hoist.
 
 
+### 📚 Libraries
+
+* grails `7.0.4 → 7.0.5`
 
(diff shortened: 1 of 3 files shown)


jskupsik (approved):

jskupsik:
I think it's good - I cleaned it up and tested it a bit in toolbox and it seems to all work.

jskupsik (approved):

PR #503, opened 2026-01-27
Tighten auth during load sequence

lbwexler:


jskupsik (approved):
Tested in toolbox with and without the hoist-react change.

PR #516, opened 2026-03-03
Add LogbackConfig.suppressStackTrace() for external loggers

lbwexler:
## Summary

- Adds `LogbackConfig.suppressStackTrace()` — a TurboFilter-based method to suppress stack traces from noisy external logger trees while preserving the exception message in log output.
- Full stack traces remain available by setting the logger level to TRACE.
- Applied to `io.micrometer.registry.otlp` in the default config.

## Test plan

- [ ] Verify that external loggers matching the prefix no longer emit stack traces at DEBUG/INFO/WARN/ERROR levels
- [ ] Verify the exception message is still appended to the log line
- [ ] Verify full stack traces appear when the logger is set to TRACE level

### DIFF
--- a/.github/workflows/deployRelease.yml
+++ b/.github/workflows/deployRelease.yml
@@ -36,7 +36,6 @@ jobs:
               fetch-tags: true
 
         - name: Validate release version and detect hotfix
-          id: validate
           env:
               VERSION: ${{ inputs.xhReleaseVersion }}
               IS_HOTFIX: ${{ inputs.isHotfix }}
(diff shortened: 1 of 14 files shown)


grails-app/init/io/xh/hoist/LogbackConfig.groovy
```diff
@@ -321,6 +329,40 @@ class LogbackConfig  {
...
+     * output. Full stack traces remain available by setting the logger level to TRACE.
+     *
+     * Note: not intended to be used for any logging done via Hoist's LogSupport, which already
+     * suppresses stack traces unless the level is TRACE.
+     */
+    protected void suppressStackTrace(String loggerPrefix) {
+        LoggerContext ctx = (LoggerContext) LoggerFactory.getILoggerFactory()
+
+        TurboFilter filter = new TurboFilter() {
+            FilterReply decide(Marker marker, Logger logger, Level level,
```
jskupsik: This 3x2 styling of parameters is neat, though I think I would prefer just a normal list of parameters. Thoughts?

PR #539, opened 2026-04-09
Auto-populate OTLP resource attributes for metrics

### LINKED ISSUE #536: Auto-populate OTLP resource attributes for metrics + allow overrides on both metrics and traces
## Problem

`TraceService` and `MetricsService` handle OTLP resource attributes inconsistently, and neither supports user overrides.

### Traces: correct defaults, no override mechanism

`TraceService.syncConfig()` builds an OTel `Resource` with four standard attributes derived from the app's runtime identity:

```groovy
// TraceService.groovy ~line 313
_resource = Resource.default.merge(
    Resource.create(Attributes.builder()
        .put(stringKey('service.name'), appCode)
        .put(stringKey('service.instance.id'), instanceName)
        .put(stringKey('deployment.environment'), appEnvironment.toString())
        .put(stringKey('service.version'), appVersion)
        .build()
    )
)
```

These values are hardcoded from Hoist's `AppEnvironment` enum, which produces values like `Development`, `Staging`, `Production`. There is no way for apps to override them via `xhTraceConfig`.

This is a real problem for platforms like DataDog, which map `deployment.environment` to their unified `env` tag. A client using DataDog has standardized on `dev` as their environment identifier, but Hoist sends `Development` — resulting in traces landing in a separate environment facet from the rest of their DD traffic.

### Metrics: no defaults at all, manual config required

`MetricsService` does not set any OTLP resource attributes automatically. Apps must manually include a `resourceAttributes` string in `xhMetricsConfig.otlpConfig` to identify themselves:

```json
{
  "otlpEnabled": true,
  "otlpConfig": {
    "url": "http://collector:4318/v1/metrics",
    "resourceAttributes": "application=someApp,env=dev"
  }
}
```

This is a Micrometer `OtlpConfig` property — a comma-separated `key=value` string — that gets passed through via the `prefixKeys()` helper to become `otlp.resourceAttributes`. Without it, metrics arrive at the collector with no service identity, no environment, and no host association. In DataDog, this means metrics show zero tags and cannot be associated with a service.

Additionally, apps configuring this manually are likely to use non-standard attribute names (e.g. `application` instead of `service.name`) since there's no guidance or framework support, further breaking the link between metrics and traces in the collector.

### The asymmetry

| Concern | TraceService | MetricsService |
|---------|-------------|----------------|
| `service.name` | Auto-set from `appCode` | Not set (must be manually configured) |
| `service.instance.id` | Auto-set from `instanceName` | Not set |
| `deployment.environment` | Auto-set from `appEnvironment.toString()` | Not set |
| `service.version` | Auto-set from `appVersion` | Not set |
| User overrides | Not supported | Supported only via raw `resourceAttributes` string |

Hoist has a strong opinion about an app's identity (`appCode`, `appEnvironment`, etc.) and should use that to populate OTLP resource attributes consistently across both signals.

### 1. Auto-populate resource attributes on MetricsService

When building the `OtlpMeterRegistry`, auto-generate the `otlp.resourceAttributes` config value from the same runtime identity TraceService uses:

```
service.name={appCode},service.instance.id={instanceName},deployment.environment={appEnvironment},service.version={appVersion}
```

This ensures metrics arrive at the collector with the same standard OTEL resource attributes as traces, with no manual config required.

### 2. Add override support to both services via soft config

Add a `resourceAttributes` map to both `xhTraceConfig` and `xhMetricsConfig` (at the top level, not nested under `otlpConfig`). User-supplied values merge with and override the auto-populated defaults:

```json
{
  "enabled": true,
  "otlpEnabled": true,
  "otlpConfig": { "endpoint": "..." },
  "resourceAttributes": {
    "deployment.environment": "dev"
  }
}
```

This allows apps to customize values (e.g. mapping `Development` → `dev` to match a platform's conventions) without losing the other defaults.

### Implementation notes

- **TraceService** (`syncConfig()`): After building the default `Resource`, merge in any entries from `config.resourceAttributes`, letting them override the defaults.
- **MetricsService** (`syncConfig()`): Build the default `resourceAttributes` string from runtime identity, then merge/override with any entries from `config.resourceAttributes`. Convert the final map to the `key=value,key=value` format Micrometer expects and inject it into the config map passed to `OtlpMeterRegistry`.
- **TraceConfig / MetricsConfig**: Add `Map resourceAttributes` property to both typed config wrappers.
- **Documentation**: Update `docs/metrics.md` and `docs/tracing.md` to document the auto-populated defaults and override mechanism.
- The `resourceAttributes` key in `otlpConfig` on the metrics side should continue to work for backwards compatibility, but the new top-level `resourceAttributes` map should take precedence when both are present. Consider logging a warning if the legacy location is used.

### LINKED ISSUE #537: Align span and metric tag names with OTEL semantic conventions
## Problem

Hoist currently uses bare, unqualified attribute names on spans and metric tags that conflict with OTEL semantic convention namespaces. This causes two issues:

1. **Namespace collisions** — bare names like `source` and `user` squat on OTEL top-level namespaces that contain their own defined attributes (e.g. `source.address`, `source.port`, `user.name`, `user.id`, `user.roles`). App developers using standard semconv attributes alongside Hoist's tags will encounter confusing conflicts.

2. **App-level collisions** — generic names like `source` are natural choices for application developers tagging their own spans (e.g. "P&L source" vs "margin source" in a data loading service). With no namespacing, app tags silently collide with Hoist's framework tags.

### Current attribute names

**On spans** (set by `TraceService.createSpan()`, ~line 149-152):
- `source` — `'app'` or `'hoist'`, classifies which layer produced the span
- `user` — the authenticated username (e.g. `jdoe`)
- `code.namespace` — the calling class name (already semconv-compliant)

**On metrics** (set by `MetricsService.applyFilters()`, ~line 165):
- `application` — the app code
- `instance` — the cluster instance name
- `source` — `'app'`, `'hoist'`, or `'infra'`

### Why these are problematic

| Current name | OTEL namespace it conflicts with | Semconv attributes in that namespace |
|---|---|---|
| `source` | `source.*` | `source.address`, `source.port` |
| `user` | `user.*` | `user.name`, `user.id`, `user.email`, `user.roles`, `user.full_name`, `user.hash` |
| `application` | — | Not a semconv attribute; duplicates `service.name` resource attribute |
| `instance` | — | Not a semconv attribute; duplicates `service.instance.id` resource attribute |

### 1. Rename span/metric tags to use semconv or `xh.` prefix

**`user` → `user.name`** on spans

The `user.*` namespace is the active, growing namespace for user identity in semconv (the older `enduser.*` namespace is being deprecated into it — `enduser.role` already redirects to `user.roles`). Hoist records the authenticated username via `HoistUser.getUsername()`, which maps directly to `user.name` ("short name or login, e.g. `a.einstein`").

**`source` → `xh.source`** on both spans and metrics

The concept of "which framework layer produced this telemetry" (`app` vs `hoist` vs `infra`) is Hoist-specific with no semconv equivalent. The `xh.` prefix is consistent with Hoist's existing namespacing conventions (e.g. `xhConfigChanged` events, `xhMetricsConfig` configs, `HOIST_ADMIN` roles).

**`deployment.environment` → `deployment.environment.name`** on resource attributes

The `deployment.environment` attribute (without `.name`) is **deprecated** as of semconv v1.27.0. Update the resource attribute set in `TraceService.syncConfig()` and the auto-populated resource attributes proposed in #536 for `MetricsService`.

**`code.namespace`** — already correct, no change needed.

### 2. Drop `application` and `instance` as metric-level tags

These are non-standard tag names that duplicate information properly conveyed via OTEL resource attributes (`service.name` and `service.instance.id`). With #536 auto-populating resource attributes on the metrics side, there is no reason to continue emitting these as metric-level tags.

This is a new telemetry system overall — there is no installed base of dashboards or queries depending on these names. Remove `application` and `instance` from the `MetricsService.applyFilters()` default tags entirely. The `xh.source` tag should remain as a metric-level tag (it has no resource attribute equivalent).

### 3. Add missing semconv attributes to SERVER spans (TraceInterceptor)

TraceInterceptor already sets several standard HTTP attributes. These additional semconv attributes are readily available from the `HttpServletRequest` and should be added:

| Semconv attribute | Stability | Hoist source | Semconv requirement |
|---|---|---|---|
| `server.port` | Stable | `request.serverPort` | Conditionally required |
| `client.address` | Stable | `request.remoteAddr` | Recommended |
| `user_agent.original` | Stable | `request.getHeader('User-Agent')` | Recommended |

### 4. Add missing semconv attributes to CLIENT spans (JSONClient)

| Semconv attribute | Stability | Hoist source | Semconv requirement |
|---|---|---|---|
| `server.port` | Stable | URI port | Conditionally required |

## Summary of all changes

| Current | Proposed | Rationale |
|---|---|---|
| `user` (span tag) | `user.name` | Semconv standard attribute, sourced from `HoistUser.getUsername()` |
| `source` (spans + metrics) | `xh.source` | Hoist-specific, avoid `source.*` namespace collision |
| `application` (metric tag) | **Remove** | Duplicates `service.name` resource attribute |
| `instance` (metric tag) | **Remove** | Duplicates `service.instance.id` resource attribute |
| `deployment.environment` (resource) | `deployment.environment.name` | Current form deprecated since semconv v1.27.0 |
| `code.namespace` (span tag) | No change | Already semconv-compliant |
| — (SERVER spans) | Add `server.port`, `client.address`, `user_agent.original` | Standard semconv attributes, data readily available |
| — (CLIENT spans) | Add `server.port` | Standard semconv attribute, data readily available |

## Future considerations

The following semconv attributes could be populated from data Hoist already has available via `HoistUser`. These are not part of the immediate scope but are worth considering in a follow-up:

**Additional user attributes:**

| Semconv attribute | Hoist source |
|---|---|
| `user.email` | `HoistUser.getEmail()` |
| `user.full_name` | `HoistUser.getDisplayName()` |
| `user.roles` | `HoistUser.getRoles()` |

**Additional SERVER span attributes:**

| Semconv attribute | Hoist source | Notes |
|---|---|---|
| `url.query` | `request.queryString` | Conditionally required per semconv |
| `network.protocol.version` | `request.protocol` | Recommended per semconv |

**Service resource attributes:**

| Semconv attribute | Potential Hoist source | Notes |
|---|---|---|
| `service.namespace` | Configurable via soft config | Groups related services composing a system |

## Background

The OTEL Semantic Conventions (currently v1.40.0) define standardized attribute names using dot-delimited namespaces. Key references:

- [Attribute naming rules](https://opentelemetry.io/docs/specs/semconv/general/naming/)
- [Attribute registry (all namespaces)](https://opentelemetry.io/docs/specs/semconv/registry/attributes/)
- [User attributes](https://opentelemetry.io/docs/specs/semconv/registry/attributes/user/)
- [Service resource attributes](https://opentelemetry.io/docs/specs/semconv/resource/service/)
- [Deployment attributes](https://opentelemetry.io/docs/specs/semconv/registry/attributes/deployment/) — note `deployment.environment` deprecated in favor of `deployment.environment.name`
- [HTTP span conventions](https://opentelemetry.io/docs/specs/semconv/http/http-spans/)

Vendor-specific mapping docs:
- [DataDog semantic mapping](https://docs.datadoghq.com/opentelemetry/mapping/semantic_mapping/)
- [DataDog unified service tagging](https://docs.datadoghq.com/getting_started/tagging/unified_service_tagging/)
- [Grafana OTLP format considerations](https://grafana.com/docs/grafana-cloud/send-data/otlp/otlp-format-considerations/)
- [AWS X-Ray exporter](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/awsxrayexporter/README.md)

Related: #536 (auto-populate OTLP resource attributes for metrics + allow overrides)

lbwexler:
## Summary

- Auto-populates `otlp.resourceAttributes` on `MetricsService` from the same standard OTEL resource attributes that `TraceService` uses (`service.name`, `service.instance.id`, `deployment.environment.name`, `service.version`), ensuring metrics and traces arrive at the collector with consistent service identity.
- Extracts the four default resource attributes into an overridable `getOtelResourceAttributes()` method on `ClusterConfig`, so apps can customize values (e.g. mapping `Development` → `dev` to match platform conventions) before services initialize.
- Both `TraceService` and `MetricsService` now consume this shared source of truth via `ClusterService.otelResourceAttributes`.

Closes #536
Closes #537

## Test plan

- [ ] Verify OTLP metrics export includes `service.name`, `service.instance.id`, `deployment.environment.name`, `service.version` resource attributes
- [ ] Verify OTLP traces continue to include the same resource attributes
- [ ] Verify overriding `getOtelResourceAttributes()` in an app's `ClusterConfig` subclass customizes both metrics and traces
- [ ] Verify existing `otlpConfig` settings (endpoint, timeout, headers) continue to work on both services

### DIFF
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -19,7 +19,7 @@ Applications can register their own custom metrics using the standard Micrometer
 - **Export registries** — built-in support for Prometheus (pull-based) and OTLP (push-based),
   configured via soft config. Additional registries (e.g. Datadog) can be added programmatically.
 - **Cluster-wide Prometheus scrape** — a single endpoint can return metrics from all instances,
-  each distinguished by an `instance` tag.
+  each distinguished by a `hoist.instance` tag.
 - **Built-in metrics** — JVM (memory, GC, threads, classloader, CPU), JDBC pool, WebSocket
   channels, client activity tracking, and Hoist monitor results are instrumented out of the box.
 - **Admin Console** — a cluster-wide metrics viewer is available via `MetricsAdminController`.
@@ -74,9 +74,9 @@ class MyService extends BaseService {
 All meters registered through the service automatically receive:
 
 1. **Default tags:**
-   - `application` — the application code (e.g. `myApp`)
-   - `instance` — the cluster instance name (e.g. `inst1`)
-   - `xh.source` — classifies the metric's origin ('hoist' or 'app')
+   - `hoist.application` — the application code (e.g. `myApp`)
+   - `hoist.instance` — the cluster instance name (e.g. `inst1`)
+   - `hoist.source` — classifies the metric's origin ('hoist' or 'app')
 
 ### Cluster-scoped metrics
 
@@ -93,7 +93,7 @@ instances.
 When `prometheusEnabled: true` in `xhMetricsConfig`, a `PrometheusMeterRegistry` is added to the
 composite registry. Prometheus scrapes are served by calling `metricsService.prometheusData()`, which
 fans out to all cluster instances via Hazelcast, collects each instance's scrape output, and
-concatenates the results. Each metric already carries an `instance` tag distinguishing its source.
+concatenates the results. Each metric already carries a `hoist.instance` tag distinguishing its source.
 
 Applications expose this via a simple controller:
 
@@ -194,7 +194,7 @@ For each configured monitor, three metrics are published:
 | `hoist.monitor.value.{code}` | Gauge | Current numeric metric value |
 | `hoist.monitor.executionTime.{code}` | Timer | Execution time of the monitor check |
 
-Each carries an `instance` tag indicating which cluster instance ran the check, or `cluster` for
+Each carries a `hoist.instance` tag indicating which cluster instance ran the check, or `cluster` for
 aggregate status. Meters are automatically removed when monitors or instances are decommissioned.
 
 See [`monitoring.md`](./monitoring.md) for full documentation of the Hoist monitoring system.
@@ -264,7 +264,7 @@ and returns a merged list of all registered meters. Each entry includes:
 - `count`, `max` — for Timer/DistributionSummary types
 - `description` — human-readable description
 - `baseUnit` — unit of measurement
-- `tags` — all tags including `application`, `instance`, `xh.source`
+- `tags` — all tags including `hoist.application`, `hoist.instance`, `hoist.source`
 - `stats` — raw statistics map
 
 This endpoint requires the `HOIST_ADMIN_READER` role.
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -75,7 +75,7 @@ All meters registered through the service automatically receive:
 
 1. **Default tags:**
    - `hoist.application` — the application code (e.g. `myApp`)
-   - `hoist.instance` — the cluster instance name (e.g. `inst1`)
+   - `hoist.instance` — the cluster instance name (e.g. `e36ca82b`)
    - `hoist.source` — classifies the metric's origin ('hoist' or 'app')
 
 ### Cluster-scoped metrics
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -15,8 +15,7 @@ Applications can register their own custom metrics using the standard Micrometer
 ### Key capabilities
 
 - **Central registry** — `MetricsService` exposes a `CompositeMeterRegistry` that all meters
-  register through. Default tags (`application`, `instance`, `source`) and namespace prefixing are
-  applied automatically.
+  register through.
 - **Export registries** — built-in support for Prometheus (pull-based) and OTLP (push-based),
   configured via soft config. Additional registries (e.g. Datadog) can be added programmatically.
 - **Cluster-wide Prometheus scrape** — a single endpoint can return metrics from all instances,
@@ -70,24 +69,14 @@ class MyService extends BaseService {
 }
 ```
 
-### Namespace prefixing and default tags
+### Default tags
 
 All meters registered through the service automatically receive:
 
 1. **Default tags:**
    - `application` — the application code (e.g. `myApp`)
    - `instance` — the cluster instance name (e.g. `inst1`)
-   - `source` — classifies the metric's origin (see below)
-
-2. **Namespace prefix** based on the `source` tag:
-   - `source=app` (default) — metric name is prefixed with the application namespace
-     (e.g. `myApp.myService.queueDepth`)
-   - `source=hoist` — prefixed with `hoist.` (e.g. `hoist.monitor.status.xhMemoryMonitor`)
-   - `source=infra` — no prefix added (e.g. `jvm.memory.used`, `jdbc.pool.active`)
-
-The namespace defaults to the application code and can be overridden via the `namespace` key in
-`xhMetricsConfig`. Note that the namespace is applied at service initialization — a restart is
-required to change it.
+   - `xh.source` — classifies the metric's origin ('hoist' or 'app')
 
 ### Cluster-scoped metrics
 
@@ -151,7 +140,7 @@ metricsService.registry.add(myDatadogRegistry)
 
 ## Built-in Metrics
 
-### JVM metrics (`source=infra`)
+### JVM metrics
 
 Automatically bound at startup via Micrometer's standard binders:
 
@@ -163,7 +152,7 @@ Automatically bound at startup via Micrometer's standard binders:
 | `jvm.classes.*` | `ClassLoaderMetrics` | Loaded and unloaded class counts |
 | `system.cpu.*` | `ProcessorMetrics` | CPU usage and available processors |
 
-### JDBC connection pool metrics (`source=infra`)
+### JDBC connection pool metrics
 
 Published by `ConnectionPoolMonitoringService` via the Tomcat JDBC pool:
 
@@ -181,7 +170,7 @@ Published by `ConnectionPoolMonitoringService` via the Tomcat JDBC pool:
 | `jdbc.pool.removeAbandoned` | Counter | Connections removed due to abandonment |
 | `jdbc.pool.releasedIdle` | Counter | Idle connections released by evictor |
 
-### WebSocket metrics (`source=infra`)
+### WebSocket metrics
 
 Published by `WebSocketService`:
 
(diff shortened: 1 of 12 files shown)


docs/metrics.md
```diff
@@ -70,24 +69,14 @@ class MyService extends BaseService {
...
-   - `source=app` (default) — metric name is prefixed with the application namespace
-     (e.g. `myApp.myService.queueDepth`)
-   - `source=hoist` — prefixed with `hoist.` (e.g. `hoist.monitor.status.xhMemoryMonitor`)
-   - `source=infra` — no prefix added (e.g. `jvm.memory.used`, `jdbc.pool.active`)
-
-The namespace defaults to the application code and can be overridden via the `namespace` key in
-`xhMetricsConfig`. Note that the namespace is applied at service initialization — a restart is
-required to change it.
+   - `hoist.application` — the application code (e.g. `myApp`)
+   - `hoist.instance` — the cluster instance name (e.g. `inst1`)
```
amcclain: Suggest: (e.g. `e36ca82b`)

grails-app/init/io/xh/hoist/ClusterConfig.groovy
```diff
@@ -82,6 +82,25 @@ class ClusterConfig {
...
+     * The returned map is used as-is — entries should use standard OTEL resource attribute keys.
+     */
+    Map<String, String> getOtelResourceAttributes() {
+        [
+            'service.name'               : appCode,
+            'service.instance.id'        : instanceName,
+            'deployment.environment.name' : appEnvironment.toString(),
+            'service.version'            : appVersion
+        ]
+    }
```
amcclain, line 102: Great, that looks like a clear plan for us 👍

grails-app/services/io/xh/hoist/telemetry/TraceService.groovy
```diff
@@ -147,9 +144,9 @@ class TraceService extends BaseService {
             pending = new SpanRef(span, span.makeCurrent(), kind)
 
         pending.setTags(tags)
-        if (!tags.source) pending.setTag('source', 'app')
+        if (!tags['hoist.source']) pending.setTag('hoist.source', 'app')
         if (caller) pending.setTag('code.namespace', caller.class.name)
-        pending.setTag('user', username ?: 'Anon')
+        pending.setTag('user.name', username ?: 'Anon')
```
amcclain: I would vote to skip the `Anon` value entirely and just not tag with a user.name - I want to avoid having to take an opinion on the placeholder string and I don't think I see the need for it

amcclain (approved):

PR #545, opened 2026-04-23
Expand typed soft-config system with nested shapes and client-side default propagation

amcclain:
## Summary

Expands the `TypedConfigMap` scheme introduced by #544 into a full typed-config system where the typed class is the single source of truth for a config's shape on both server and client. Builds directly on Lee's `init(Map)` ordering fix from #544 — does not revisit or revert that choice.

## What the two commits do

**Infrastructure + migration**

- `TypedConfigMap` now `implements JSONFormat`; `formatForJSON()` returns a shallow Map and Jackson's `JSONFormatSerializer` handles recursion at render time (no hand-rolled serializer).
- Adds nested-typed and `List<TypedConfigMap>` recursion on the apply path, plus an optional `getConfigName()` for top-level classes.
- New `ConfigService.getTypedConfig(Class)` API; new optional `typedClass:` key on `ensureRequiredConfigsCreated` entries that registers the typed class, validates `getConfigName()` alignment, and logs a startup WARN when typed-class property defaults diverge from BootStrap `defaultValue` (recurses through nested typed shapes and typed lists).
- 12 new typed classes (AlertBanner, ActivityTracking+nested, Changelog, ClientError, ConnPoolMonitoring, EnvPoll, Export, Idle, Ldap+nested server list, LogArchive, MemoryMonitoring, WebSocket); the 3 existing ones (Monitor, Metrics, Trace) gain defaults + per-key Groovydoc.
- All 15 built-in hoist-core JSON configs wired through `typedClass:` in BootStrap.
- 11 server-side call sites migrated from `configService.getMap(name)` → `configService.getTypedConfig(Class)`, dropping inline `?:` fallbacks and `!= false` patterns where the typed class now supplies the default.
- Several latent config edge-cases tightened up during migration (schema gaps, missing defaults, defensive coercions).
- Docs: `docs/configuration.md` typed-config section, `CLAUDE.md` new Groovy-conventions bullet on omitting `public` modifiers (plus the field-vs-method nuance), `CHANGELOG.md`, `doc-registry.json`.

**Client-bound sanitation parity**

- `getClientConfig()` now routes typed configs through `config.externalValue(obscurePassword: true, jsonAsObject: true)` as the uniform sanitation step, then populates the typed class from that sanitized Map. Closes an architectural gap where the typed branch was using server-side opts (`decryptPassword: true`) instead of the client-side opts used everywhere else in the method. No behavioral change today for JSON configs; future-proofs against new sanitation opts added to `externalValue`.

## Alignment with #544

- Lee's `init(Map)` pattern is the authoritative apply method. Each new typed class declares a single `MyConfig(Map args) { init(args) }` constructor — no `super(args)` anywhere.
- Lee's WARN-on-unknown-keys semantics preserved (louder than the original DEBUG plan; better for catching typos).
- Tracing refactor respected: new `TraceConfig` lives at `io.xh.hoist.telemetry.trace.TraceConfig`, `MetricsConfig` at `io.xh.hoist.telemetry.metric.MetricsConfig`.
- One small additional fix shipped alongside: `instanceLog.logWarn(...)` → `logWarn(...)` in `TypedConfigMap.init()`. The previous receiver was a plain SLF4J `Logger` which has no `logWarn` method; unknown-key path would have thrown `MissingMethodException` on first hit. Flagged this directly to Lee; included here for convenience unless Lee prefers to own it separately.

## Not in this PR (deferred)

- Nesting `TraceConfig.otlpConfig` into a typed `OtlpTraceOptions` class. Lee just rewrote TraceService; holding until the tracing dust settles.

## Test plan

- [ ] `./gradlew assemble` green (CI).
- [ ] Boot a downstream app against this branch; verify all 15 `xh*` configs still load cleanly and admin-console edits round-trip.
- [ ] Confirm `getClientConfig()` payload shape matches pre-change output for each typed config.
- [ ] Confirm no divergence-WARN on startup for unchanged default sets (spot-check: MonitorConfig, LdapConfig, ActivityTrackingConfig).
- [ ] Drop a config row from DB and restart — verify typed-class defaults populate correctly.
- [ ] Typo an admin JSON value (add an unknown key) — verify WARN logs with the subclass + config name context.

---

Pairs with a one-line hoist-react change (`xhAppInstances` empty default in `DifferModel.ts`) which was taken directly to `develop`.

🤖 Generated with Claude Opus 4.7 (1M context)


grails-app/services/io/xh/hoist/admin/ConnectionPoolMonitoringService.groovy
```diff
@@ -79,12 +79,12 @@ class ConnectionPoolMonitoringService extends BaseService {
 
         // Don't allow snapshot history to grow endlessly -
         // default cap @ 1440 samples, i.e. 24 hours * 60 snaps/hour
-        if (_snapshots.size() > (config.maxSnapshots ?: 1440)) {
+        if (_snapshots.size() > config.maxSnapshots) {
             def oldest = _snapshots.min {it.key}
             _snapshots.remove(oldest.key)
```
lbwexler, line 84: nice!

grails-app/services/io/xh/hoist/config/ConfigService.groovy
```diff
@@ -33,6 +33,12 @@ import static io.xh.hoist.json.JSONSerializer.serializePretty
 @GrailsCompileStatic
 class ConfigService extends BaseService {
 
+    // Registry of typed config classes, keyed by backing config name. Populated when apps declare
+    // a `typedClass:` entry in `ensureRequiredConfigsCreated`. Used to (a) validate name/class
+    // alignment at startup, (b) populate `getClientConfig()` payloads with typed-class defaults,
+    // and (c) flag drift between declared property defaults and BootStrap `defaultValue`.
```
lbwexler: do we still need/want BootStrap defaultValue?

grails-app/services/io/xh/hoist/config/ConfigService.groovy
```diff
@@ -83,7 +114,19 @@ class ConfigService extends BaseService {
...
             def name = config.name
             try {
-                ret[name] = config.externalValue(obscurePassword: true, jsonAsObject: true)
+                // Sanitize via `externalValue` as the single source of truth for client-bound
+                // output — handles instance-config overrides, JSON parsing, and password
+                // obscuring uniformly for every config.
+                def external = config.externalValue(obscurePassword: true, jsonAsObject: true)
+                def typedClass = typedConfigs[name]
+                if (typedClass && config.valueType == 'json' && external instanceof Map) {
+                    // Populate the typed class from the sanitized external value. Typed instance
```
lbwexler: would we want formatForJson -- or a form of it -- e.g. formatForClientJson -- allow us to selectively send certain properties to the client -- we often need that and the typed map is a good place to do it.

grails-app/services/io/xh/hoist/config/ConfigService.groovy
amcclain: I think we could use it more judiciously - advantage is that it fills in the shape for you, so you have more guidance as to what can go in there.  But clearly just waiting to fall out of date and eagerly-inserting the initial defaults at config creation time, never updating them as config shape changes....  

At some point I think you are right - we remove them and then maybe look to drive better hints to user in UI- the new shape class hopefully sets us up with more options there if / when we want to look at it.

grails-app/services/io/xh/hoist/config/ConfigService.groovy
```diff
@@ -65,6 +71,31 @@ class ConfigService extends BaseService {
...
+    /**
+     * Load a typed representation of a JSON soft config, with declared property defaults
+     * applied for any keys missing from the stored value.
+     *
+     * The supplied class must extend {@link TypedConfigMap} and declare the backing config
+     * name via `getConfigName()`. This is the preferred way to read structured configs —
+     * it centralizes defaults and documentation on the typed class itself, rather than
+     * scattering `?:` fallbacks across call sites.
+     */
+    <T extends TypedConfigMap> T getTypedConfig(Class<T> clazz) {
```
lbwexler: getObject() ? simpler and more inline with the other gettters?

src/main/groovy/io/xh/hoist/admin/ConnPoolMonitoringConfig.groovy
```diff
@@ -0,0 +1,32 @@
...
+
+import io.xh.hoist.config.TypedConfigMap
+
+/**
+ * Typed representation of the `xhConnPoolMonitoringConfig` soft config, governing the
+ * `ConnectionPoolMonitoringService` that periodically samples JDBC pool state.
+ */
+class ConnPoolMonitoringConfig extends TypedConfigMap {
+
+    String getConfigName() { 'xhConnPoolMonitoringConfig' }
```
lbwexler: Sure you considered, but it would be great if this could just be a property initializer.  Also could consider passing to the constructor.  

Curious to understand what this is needed for -- this does prevent us from re-suing this object for multiple configss.

amcclain: This lets us access the typed config from ConfigService w/o having to pass type class + config name in those calls - a feature from the POV of not potentially getting the config name wrong, removed what looked like some duplication.  Did not consider reusing shape across different configs - not a use case that came to my mind but am sympathetic to the possibility

src/main/groovy/io/xh/hoist/track/ActivityTrackingConfig.groovy
```diff
@@ -0,0 +1,112 @@
...
+
+    /** Nested options for client health reports. */
+    static class ClientHealthReport extends TypedConfigMap {
+        /** Interval (minutes) between client health-check track entries. -1 disables. */
+        Integer intervalMins = -1
+
+        ClientHealthReport(Map args) { init(args) }
+    }
+
+    /**
```
lbwexler, line 76: build this workaround into the super class?  just a thought.

grails-app/services/io/xh/hoist/config/ConfigService.groovy
```diff
@@ -201,11 +261,105 @@ class ConfigService extends BaseService {
...
         }
 
         logDebug("Validated presense of ${reqConfigs.size()} required configs", "created ${created}")
     }
 
+    //-------------------
+    //  Typed Config Registration
+    //-------------------
+    private void registerTypedConfig(String confName, Class typedClass, Map bootstrapDefault) {
+        if (!TypedConfigMap.isAssignableFrom(typedClass)) {
```
lbwexler, line 265: throw?

grails-app/services/io/xh/hoist/config/ConfigService.groovy
```diff
@@ -65,6 +71,31 @@ class ConfigService extends BaseService {
...
+     * The supplied class must extend {@link TypedConfigMap} and declare the backing config
+     * name via `getConfigName()`. This is the preferred way to read structured configs —
+     * it centralizes defaults and documentation on the typed class itself, rather than
+     * scattering `?:` fallbacks across call sites.
+     */
+    <T extends TypedConfigMap> T getTypedConfig(Class<T> clazz) {
+        // Resolve config name: prefer registry (avoids constructing twice), fall back to an
+        // empty instance to read the name if the class hasn't been registered via BootStrap.
+        String name = typedConfigs.find { it.value == clazz }?.key
+            ?: ((TypedConfigMap) clazz.getDeclaredConstructor(Map).newInstance([:])).configName
```
lbwexler: why not just require that it be registered, actually  This would allow us to do without the configName

amcclain:
FYI — once this lands, the [`Use configService Typed Getters`](https://github.com/xh/hoist-core/blob/develop/docs/coding-conventions.md#use-configservice-typed-getters) section of `docs/coding-conventions.md` should be updated to introduce `configService.getTypedConfig(Class)` as the recommended path for stable-shape JSON configs. The current section only covers the legacy typed getters (`getMap`, `getString`, etc.).

lbwexler:
will do. turning to this right now

amcclain:
(I setup that coding-conventions doc over the weekend as part of some MCP work I was doing here.  Had CC push comments to this and the other PR before I forgot to ensure we get things updated once these land)

PR #547, opened 2026-04-24
Add tailSamplingEnabled flag to xhTraceConfig

amcclain:
## Summary

Proposes adding a `tailSamplingEnabled` flag to `xhTraceConfig` as an opt-out for the tail-sampling buffer introduced on the parent branch. When set to `false`, `SpanProcessingService` hands each completed server and client span directly to the export pipeline as it finishes, and `sampleRate` / `sampleRules` / `traceTimeoutMs` / `maxBufferedTraces` are ignored. Sampling is then effectively delegated to whatever infrastructure the operator runs downstream (e.g. an OTel Collector).

Defaults to `true`, preserving the current branch's behavior exactly.

## Motivation

A couple of operational modes currently available through `xhTraceConfig` already approximate pass-through by setting `sampleRate: 1.0` with no rules — every trace ends up kept — but spans still accumulate in the per-trace buffer until the root ends, and the `MAX_SPANS_PER_TRACE` (500) and `maxBufferedTraces` (10k) caps still apply. For deployments that want the simpler "record and export as you go" shape and rely on a collector for keep/drop decisions, this flag makes that an explicit first-class mode rather than an emergent side effect of the config.

It also leaves a clean, well-labeled seam: operators who don't want Hoist making sampling decisions at all can disable it in one knob, without losing any of the other instrumentation (`onStart` attribute stamping, W3C propagation, JDBC tracing, OTLP export, client-span relay).

## What changes

- **`TraceConfig`**: new `boolean tailSamplingEnabled = true` field.
- **`BootStrap`**: new key added to the default `xhTraceConfig` value.
- **`SpanProcessingService`**:
  - `onStart` still stamps common attributes in both modes.
  - `onEnd` short-circuits to `traceService.exportSpans([span])` when the flag is off.
  - `submitClientSpans` short-circuits the same way for browser-relayed spans.
  - `sweep` drains any leftover buffers unconditionally on a live flip from `true` → `false`, rather than rolling them against the now-superseded sample rate.
- **`docs/tracing.md`**: new config row, new "Pass-through mode" subsection under Sampling.

## What stays the same

- `ParentBased(AlwaysOn)` sampler is unchanged. Inbound W3C `sampled=0` still drops locally, and outbound `traceparent` still carries `sampled=1` while recording.
- All span attribute stamping (`user.name`, `xh.isPrimary`, `xh.impersonating`, `xh.fromHoistClient`, HTTP semantic-convention attrs from `HoistFilter`) runs in both modes.
- `ClientSpanData` / client span relay still merges client and server spans by traceId into a single trace in the backend (via direct export in pass-through mode, via the tail buffer otherwise).
- `BatchSpanProcessor` handling of each registered exporter is unchanged.

## Trade-offs worth calling out for review

- **Default value.** Chose `true` to preserve current branch behavior exactly. If we'd rather make the simpler mode the default, it's a one-line change.
- **Live flip behavior.** Flipping `true` → `false` drains any in-flight buffers on the next `sweep` tick (≤5s) and exports them unconditionally. Flipping `false` → `true` just starts buffering new traces from that point; no retroactive behavior.
- **Admin stats.** `bufferedTraces` reads `0` when the flag is off — self-documenting, no extra field needed.

## Test plan

- [ ] Enable tracing with `tailSamplingEnabled: true` (default); confirm existing behavior unchanged — traces buffer and flush on root end, error traces kept in full, sample rules applied at root.
- [ ] Flip to `tailSamplingEnabled: false` at runtime; confirm `bufferedTraces` drains within one sweep interval and new spans export individually as they complete.
- [ ] In pass-through mode, confirm browser-submitted spans still appear alongside their server counterparts in the backend (same traceId).
- [ ] In pass-through mode, confirm inbound `traceparent` with `sampled=0` still drops locally (no spans recorded).
- [ ] In pass-through mode, confirm outbound `JSONClient` calls still carry `traceparent` with `sampled=1`.

---
_Generated by [Claude Code](https://claude.ai/code/session_01B694STHxPV4xurGXeFotAe)_

### DIFF
--- a/docs/tracing.md
+++ b/docs/tracing.md
@@ -223,6 +223,7 @@ ObservedRun.observe(this)
 ```json
 {
     "enabled": false,
+    "tailSamplingEnabled": true,
     "sampleRate": 1.0,
     "sampleRules": [],
     "traceTimeoutMs": 300000,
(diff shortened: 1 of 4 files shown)


amcclain:
Lmk if this seems worthwhile? Happy to discuss when we have a chance

lbwexler:
Thanks for the excellent AI assisted review, and the suggestion of the flag, which really helped point me to understanding our story here better, and how tail sampling was actually just a misstep.

We'll go with a much more appropriate, standards-based head sampling that can easily be turned off in favor of proper production tail sampling at the collector level.  I am solving the problem with the complicated and partially implemented alwaysSampleErrors by just removing it.

PR #554, opened 2026-05-05
Add service.build.id to OTLP resource attributes

lbwexler:
## Summary

Adds `service.build.id` (from `Utils.appBuild`) to the resource attributes returned by `ClusterConfig.getOtelResourceAttributes()`, alongside the existing `service.version`. Omitted when the build is unset or `'UNKNOWN'`. 

## Followup worth considering (separate issue)

While making this change we noticed our spans aren't really set up to surface client/server version mismatches. Today the server unilaterally tags `service.version` (and now `service.build.id`) based on its own build, with nothing on the span linking back to what version of hoist-react / app client actually originated the request. That makes it hard to spot or debug version-skew issues from telemetry alone. Worth tracking as its own issue.

## Test plan

- [ ] Verify OTLP exports include `service.build.id` when `appBuild` is set to a real value
- [ ] Verify the attribute is omitted when `appBuild` is `'UNKNOWN'` or null

### DIFF
--- a/.mcp.json
+++ b/.mcp.json
@@ -20,7 +20,7 @@
       }
     },
     "jetbrains": {
-      "url": "http://localhost:64342/sse",
+      "url": "http://localhost:${JETBRAINS_MCP_PORT:-64342}/sse",
       "type": "sse"
     }
   }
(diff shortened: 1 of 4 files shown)


amcclain (approved):
Great thank you

PR #556, opened 2026-05-06
Configurable Timers/Counters and client-side metrics ingestion

lbwexler:
(See also https://github.com/xh/toolbox/pull/850)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,24 @@
 
 ## 40.0-SNAPSHOT - unreleased
 
+### ⚠️ Breaking Changes (minor)
+* Removed `ObservedRun.timer(Timer)` and `ObservedRun.counter(Counter)` - the pre-built-instance
+  variants. Use the by-name forms `timer(name: ...)` / `counter(name: ...)` and configure
+  Timer-level options centrally via `MetricsService.configureTimer`.
+* `ObservedRun.counter(name: ...)` semantic changed: previously incremented before the closure
+  ran (counted attempts), now increments on completion alongside the timer. Both timer and
+  counter now attach an `xh.outcome` tag with value `success` or `failure` based on whether
+  the closure threw.
+
+
 ### 🎁 New Features
 
+* Client-side metrics support: new `/xh/recordMetrics` endpoint accepts batched timer and counter
+  entries from the JS client.
+* New `MetricsService.configureTimer(name, ...)` and `configureCounter(name, ...)` for
+  one-shot init-time configuration of named meters. `configureTimer` sets percentiles, SLOs,
+  and histogram bounds applied to all tagged variants of the configured name; both surface a
+  description in the admin metrics view.
 * OTLP exports now include a `service.build.id` resource attribute alongside `service.version`.
 
 ### 🤖 AI Docs + Tooling
@@ -15,6 +31,8 @@
 * `hoist-core-symbols` now indexes `.java` sources alongside `.groovy` under
   `src/main/groovy/`, surfacing types previously missing from search and member listings
   (e.g. `JSONSerializer`, `JSONParser`, `JSONFormat`, `CollectionUtils`, `RoutineException`).
+
+### 🐞 Bug Fixes
 
 * Fixed `StandardMetricsService` JDBC pool metrics throwing NPE when JDBC tracing is enabled,
   and hardened `MetricsAdminService.listMetrics()` so a single throwing meter no longer breaks
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,10 +16,13 @@
 
 * Client-side metrics support: new `/xh/recordMetrics` endpoint accepts batched timer and counter
   entries from the JS client.
-* New `MetricsService.configureTimer(name, ...)` and `configureCounter(name, ...)` for
+* New `BaseService.configureTimer(name, ...)` and `BaseService.configureCounter(name, ...)` for
   one-shot init-time configuration of named meters. `configureTimer` sets percentiles, SLOs,
   and histogram bounds applied to all tagged variants of the configured name; both surface a
   description in the admin metrics view.
+* New `BaseService.telemetryPrefix` property — when set on a subclass, is auto-prepended (with a
+  `.` separator) to metric and span names registered via `BaseService.configureMetricTimer` /
+  `configureMetricCounter` and `ObservedRun.span` / `.timer` / `.counter`.
 * OTLP exports now include a `service.build.id` resource attribute alongside `service.version`.
 * `RestController` `update` now also accepts `PATCH` (in addition to `PUT`).
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,21 +6,23 @@
 
 * Removed `ObservedRun.timer(Timer)` and `ObservedRun.counter(Counter)` - the pre-built-instance
   variants. Use the by-name forms `timer(name: ...)` / `counter(name: ...)` and configure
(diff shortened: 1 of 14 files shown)


grails-app/services/io/xh/hoist/telemetry/metric/impl/BuiltInMetricsService.groovy
```diff
@@ -40,7 +41,7 @@ import javax.sql.DataSource
  * @internal - not intended for direct use by applications.
  */
 @CompileStatic
-class StandardMetricsService extends BaseService {
+class BuiltInMetricsService extends BaseService {
```
amcclain, line 42: I'm a-OK with this name as-is. 

Checking if `HoistMetricsService` would make sense since these all get `source:hoist`? Also thought of `DefaultMetricsService` but I don't think that's the right usage of `Default` relative to our other usages - i.e. `DefaultRoleService` is a service apps can use directly for their own roles, which is not what this is. Only asking as `BuiltIn` is a new prefix for us.

CHANGELOG.md
```diff
@@ -16,6 +32,8 @@
   `src/main/groovy/`, surfacing types previously missing from search and member listings
   (e.g. `JSONSerializer`, `JSONParser`, `JSONFormat`, `CollectionUtils`, `RoutineException`).
 
+### 🐞 Bug Fixes
+
 * Fixed `StandardMetricsService` JDBC pool metrics throwing NPE when JDBC tracing is enabled,
```
amcclain, line 43: Update to new name?

grails-app/services/io/xh/hoist/telemetry/metric/MetricsService.groovy
```diff
@@ -144,6 +269,21 @@ class MetricsService extends BaseService {
...
+        registry.config().meterFilter(new MeterFilter() {
+            DistributionStatisticConfig configure(Meter.Id id, DistributionStatisticConfig config) {
+                def spec = _timerSpecs[id.name]
+                if (!spec) return config
+                def b = DistributionStatisticConfig.builder()
+                if (spec.percentiles) b.percentiles(spec.percentiles as double[])
+                if (spec.slos) b.serviceLevelObjectives(spec.slos.collect { it.toNanos() as double } as double[])
+                if (spec.publishHistogram) b.percentilesHistogram(true)
+                if (spec.minExpected) b.minimumExpectedValue(spec.minExpected.toNanos() as Double)
+                if (spec.maxExpected) b.maximumExpectedValue(spec.maxExpected.toNanos() as Double)
```
amcclain: Truthiness seems problematic here - assume `0` could be a valid value

grails-app/services/io/xh/hoist/telemetry/metric/MetricsService.groovy
```diff
@@ -183,7 +323,7 @@ class MetricsService extends BaseService {
     }
 
     private static boolean isDefaultHoistSource(String name) {
```
amcclain, line 490: Funny that we have both `default` and `hoist` in the name here 😄

grails-app/services/io/xh/hoist/telemetry/metric/MetricsService.groovy
```diff
@@ -76,12 +84,78 @@ class MetricsService extends BaseService {
...
     void init() {
         registry.add(readOnlyRegistry)
         applyFilters()
         syncConfig()
     }
 
+    /**
+     * Configure a named Timer with distribution statistics and metadata.
+     *
+     * Apply once at app init - typically from your app's metrics service or bootstrap.
```
amcclain: Would this often make the most sense to do in the init of the particular service where the Timer is used?  That will effectively be called from Bootstrap, but this guidance makes it sound like it's actively important to locate these in a central place, outside of individual business services.

grails-app/services/io/xh/hoist/telemetry/metric/MetricsService.groovy
```diff
@@ -76,12 +84,78 @@ class MetricsService extends BaseService {
...
+     *
+     * Distribution config (percentiles, slos, etc.) is applied via a {@link MeterFilter} and
+     * must be configured before the affected meters are first registered, otherwise the meter
+     * is created with the unfiltered config and this call has no effect for that instance.
+     *
+     * Description is stored in a side-map and surfaced through the admin metrics view; it is
+     * not exported through Micrometer's {@code Meter.Id} description (which would require
+     * pre-registering an anchor meter per name).
+     *
+     * @param name              metric name (must match the name used at the call site)
```
amcclain, line 115: I don't think we need to do this now, but this might be the most natural time to automate the prepending off an appCode prefix to all metric names - so I could get that automatically without having to spec e.g. `toolbox.fooMetric` both where I am defining and then using it.  

But I don't see any obvious way to support this without getting in the way of Hoist or app-plugin usages of these same service APIs, other than piling more conventions on top.... 

I'm not attached to this, just throwing it out there to note we've considered it, if this sparks some clear idea on how to support, great, otherwise it can continue to be responsibility of app devs to manage their namespacing.

grails-app/services/io/xh/hoist/telemetry/metric/MetricsService.groovy
```diff
@@ -76,12 +84,78 @@ class MetricsService extends BaseService {
...
+     * @param description       human-readable description shown in the admin metrics view
+     * @param percentiles       client-side percentiles to compute (e.g. [0.5, 0.95, 0.99])
+     * @param slos              service-level-objective buckets for histogram output
```
amcclain, line 121: Interesting that SLOs can be defined at this level here - I had assumed those were something that were setup more in the monitoring platform that consumes the feed (DataDog, etc).  I assume that this optional param doesn't limit us from that in any new way - just noting that SLOs are on my list to understand better all around, and we might then have an app / enterprise level decision around if they should be set here or in monitoring tools.

amcclain (commented):
Looks great, thanks for keeping this continuous stream of enhancements flowing.

One potential bug concern re. truthiness, otherwise I think my comments are mostly just that - general notes or "what about..." thoughts we don't need to act on now.

amcclain:
Pushed the catch-up merge with `develop` to confirm that we have CI passing.

grails-app/services/io/xh/hoist/telemetry/metric/impl/BuiltInMetricsService.groovy
lbwexler: Don't feel strongly about the name -- just wanted to avoid confusing "users" looking at the API for the public surface -- hence the move to "impl" and the less appealing name

grails-app/services/io/xh/hoist/telemetry/metric/MetricsService.groovy
```diff
@@ -76,12 +87,122 @@ class MetricsService extends BaseService {
...
+     *                          call site's tag set
+     * @param percentiles       client-side percentiles to compute (e.g. [0.5, 0.95, 0.99])
+     * @param slos              service-level-objective buckets for histogram output
+     * @param publishHistogram  publish a percentile histogram for server-side aggregation
+     *                          (e.g. Prometheus histogram_quantile)
+     * @param minExpected       minimum expected value - used to size the histogram
+     * @param maxExpected       maximum expected value - used to size the histogram
+     */
+    @NamedVariant
+    void createTimer(
```
amcclain: This doesn't compile - clashes with `BaseService.createTimer(Map)`.  Can we go back to `configureTimer` / `configureCounter` here?

PR #557, opened 2026-05-06
Fix JSONClient timeout so requests actually abort on timeout (#241)

### LINKED ISSUE #241: fix http client timeout so that requests really do abort upon timeout
We noticed while looking at client logs, that requests that are setup with 30s or 1mn timeouts can sometimes not abort/hangup once the timeout is reached.   We've seen requests last up to ~10mn.

As explained in this post
https://www.baeldung.com/httpclient-timeout
our current timeout configs do not hard abort a request.

The consequences are that under heavy load and slow response times from a remote host, our server is not freeing up request threads as fast as we thought, and performance is degrading.

To get real request aborts, we want to add a `TimerTask` with a `method.abort() `call to the `method` passed into https://github.com/xh/hoist-core/blob/develop/src/main/groovy/io/xh/hoist/http/JSONClient.groovy#L145

lbwexler:
## Summary

Closes #241.

`JSONClient` now exposes a single `timeoutMs` that bounds the total duration of every request issued by the client — settable per-client (constructor) or overridden per-call (optional method arg). When exceeded, the in-flight request is forcibly aborted and surfaces as `ExternalHttpException` with `SC_GATEWAY_TIMEOUT` (504).

This replaces the earlier approach of reading `RequestConfig.responseTimeout` off the request and computing a backstop from HttpClient's per-phase knobs. Apache HttpClient's per-phase timeouts (`connect` / `connectionRequest` / `response`) provide no end-to-end deadline and have documented gaps for proxy CONNECT, TLS handshake, slow-trickle response bodies, and retry multiplication. A single explicit abort scheduled at `now + timeoutMs` is the simple guarantee callers actually want.

## API

```groovy
// Constructor — named or positional via @NamedVariant
new JSONClient()
new JSONClient(customClient)
new JSONClient(customClient, 5000)
new JSONClient(timeoutMs: 5000)
new JSONClient(client: customClient, timeoutMs: 5000)

// Per-call override
client.executeAsMap(request)              // uses client default
client.executeAsMap(request, 1000)        // override
```

`timeoutMs` is `Long` — null (default) disables the timeout.

## Notes

* Single shared static daemon `Timer` for scheduling aborts. Abort tasks are non-blocking so the single-thread-Timer caveat doesn't apply.
* Convenience methods use try-with-resources composing `AbortHandle` (a small AutoCloseable) and `CloseableHttpResponse`. Replaces the previous `try/finally { abortTask?.cancel() }` shape.
* `AbortHandle.fired` is set only by *our* scheduled task — so third-party `method.abort()` calls don't get mislabeled as our timeout.
* Callers who do not configure `timeoutMs` see no behavior change (null = disabled).

## Test plan

- [ ] Manual: hit a deliberately stalled endpoint with `new JSONClient(timeoutMs: 100)`; confirm the request aborts at deadline and surfaces as 504 `ExternalHttpException`.
- [ ] Manual: confirm normal successful + error-status responses are unchanged.
- [ ] Manual: confirm callers using `new JSONClient()` with no timeout see no behavior change.
- [ ] Manual: confirm per-call override (`executeAsMap(method, 50)`) takes precedence over the client's default.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,9 +19,8 @@
 * Fixed `StandardMetricsService` JDBC pool metrics throwing NPE when JDBC tracing is enabled,
   and hardened `MetricsAdminService.listMetrics()` so a single throwing meter no longer breaks
   the admin metrics view.
-* `JSONClient` now schedules a hard `abort()` backstop on requests that have a bounded
-  `responseTimeout`, ensuring blocked socket reads are released and surfaced as a 504-flavored
-  `ExternalHttpException` rather than hanging request threads (#241).
+* `JSONClient` now hard-aborts requests that exceed their `responseTimeout`, surfacing as a
+  504 `ExternalHttpException` instead of hanging the request thread (#241).
 
 
 ## 39.0.1 - 2026-04-30
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,9 @@
 * Fixed `StandardMetricsService` JDBC pool metrics throwing NPE when JDBC tracing is enabled,
   and hardened `MetricsAdminService.listMetrics()` so a single throwing meter no longer breaks
   the admin metrics view.
+* `JSONClient` now schedules a hard `abort()` backstop on requests that have a bounded
+  `responseTimeout`, ensuring blocked socket reads are released and surfaced as a 504-flavored
+  `ExternalHttpException` rather than hanging request threads (#241).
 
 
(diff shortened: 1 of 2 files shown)


cnrudd (commented):
All looks good except it's not clear to me why `response?.close()` is no longer needed.

cnrudd:
> All looks good except it's not clear to me why `response?.close()` is no longer needed.

@lbwexler explained: try with resources closes a closeable.

PR #558, opened 2026-05-07
Merge user-comment supplements onto prior client error rows (#352)

### LINKED ISSUE #352: Improve sending of emails from handleException
When `handleException logOnServer: true`,  and a user sends the error with a message/feedback, hoist-core ends up sending 2 emails:

1: first email sent without error content (was triggered when handleException dialog was opened)
2: second email sent without error content and without user's message/feedback  (this one triggered user's hitting of Send button with a message)

This can create some confusion for support teams reading these incoming emails.  They have to know to go to the App's Admin Activity Errors panel to read the error content and the user's message.

lbwexler:
Closes #352.

When the exception dialog auto-logs a client error, then the user submits a comment via the same dialog, hoist-react makes two `track()` calls — one without `userMessage`, one with. Result: two TrackLog rows for one logical error, and `ClientErrorEmailService` sends two emails (the first lacking the user's message).

This change detects the supplement in `TrackService.trackAll`: when an incoming Client Error entry carries a `userMessage` and a recent (15m) prior row exists for the **same user, same loadId**, with byte-identical `data.error` and no existing `userMessage`, the comment is merged onto the prior row and a new `xhClientErrorCommented` cluster topic is published. `ClientErrorEmailService` subscribes (primary-only) and sends a single "User Comment Added" follow-up email.

## Match key

- `category == 'Client Error'`
- same `username`
- same `loadId` (so we are guaranteed it is the same browser tab session)
- `dateCreated` within last 15 minutes
- prior row has no `userMessage`
- prior row's `data.error` is structurally equal to the incoming `data.error` (Groovy Map equality on the sanitized exception payload — byte-identical because both calls go through `sanitizeException` on the same `HoistException` instance).

Bounded to 10 candidate rows. Anything missing → fall through to the existing path and create a new row, no behavior change.

## Test plan

- [x] Trigger an error with `logOnServer: true`, dismiss the dialog without commenting → one TrackLog row, one email (existing behavior).
- [x] Trigger an error with `logOnServer: true`, type a comment and Send → one TrackLog row with `userMessage` populated, two emails (the original digest + a "User Comment Added" follow-up).
- [x] Refresh the tab between auto-log and Send (different `loadId`) → two rows, no merge (correct, comment is a different session).
- [x] Two distinct errors in the same session, both commented → two rows, two follow-up emails.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 * OTLP exports now include a `service.build.id` resource attribute alongside `service.version`.
 * `RestController` `update` now also accepts `PATCH` (in addition to `PUT`).
+* When a user submits a comment via the exception dialog, `TrackService` now merges it onto the
+  matching prior auto-logged client error (same user, same `data.error`, within 30m) instead of
+  creating a duplicate row. `ClientErrorEmailService` sends a follow-up "User Comment Added"
+  email on the new `xhClientErrorCommented` topic.
 
 ### 🤖 AI Docs + Tooling
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,10 +6,11 @@
 
 * OTLP exports now include a `service.build.id` resource attribute alongside `service.version`.
 * `RestController` `update` now also accepts `PATCH` (in addition to `PUT`).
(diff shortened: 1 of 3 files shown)


cnrudd:
tested in toolbox.  works well.

PR #562, opened 2026-05-14
Grails 7.1 upgrade

lbwexler:
## Summary
- Upgrade to Grails 7.1 (Spring Boot 3.5, Hibernate 5).
- Upgrade Open Telmetry, Hazelcast and other Misc.
- Fix `PooledDataSource` unwrapping so JDBC metrics + pool monitoring continue to work when `JdbcTraceService` wraps the chain with `OpenTelemetryDataSource`. New shared `DataSourceUtils.findPooledDataSource()` walks past Spring's `DelegatingDataSource` chain and reflects through `OpenTelemetryDataSource.delegate`. Resolution is eager in both services for consistency.

## Test plan
- [ ] App boots cleanly with no `Primary DataSource not org.apache.tomcat.jdbc.pool.DataSource` warning
- [ ] `jdbc.connections.*` gauges and counters appear in metrics
- [ ] Admin console connection pool monitoring shows live snapshots
- [ ] Behavior verified with JDBC tracing both enabled and disabled

### DIFF
--- a/grails-app/services/io/xh/hoist/admin/ConnectionPoolMonitoringService.groovy
+++ b/grails-app/services/io/xh/hoist/admin/ConnectionPoolMonitoringService.groovy
@@ -8,14 +8,15 @@
 package io.xh.hoist.admin
 
 import io.xh.hoist.BaseService
+import io.xh.hoist.config.ConfigService
 import io.xh.hoist.exception.DataNotAvailableException
 import org.apache.tomcat.jdbc.pool.DataSource as PooledDataSource
 import org.apache.tomcat.jdbc.pool.PoolConfiguration
-import org.springframework.boot.jdbc.DataSourceUnwrapper
 
 import javax.sql.DataSource
 import java.util.concurrent.ConcurrentHashMap
 
+import static io.xh.hoist.util.DataSourceUtils.findPooledDataSource
 import static io.xh.hoist.util.DateTimeUtils.SECONDS
 import static io.xh.hoist.util.DateTimeUtils.getHOURS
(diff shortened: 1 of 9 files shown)


amcclain (approved):

grails-app/services/io/xh/hoist/admin/ConnectionPoolMonitoringService.groovy
```diff
@@ -28,15 +29,18 @@ import static java.lang.System.currentTimeMillis
...
     private Map<Long, Map> _snapshots = new ConcurrentHashMap()
     private Date _lastInfoLogged
-    private PooledDataSource _pooledDataSource
-    private boolean unwrapAttempted = false
+    private PooledDataSource pooledDataSource
 
     void init() {
+        pooledDataSource = findPooledDataSource(dataSource)
+        if (!pooledDataSource) {
+            logError("Failed to unwrap primary Datasource to org.apache.tomcat.jdbc.pool.DataSource - cannot monitor connection pool usage.")
```
jskupsik, line 42: Is Error the correct level for this? It was previously a warn. Also I don't think this is wrapped behind any config so it will log error even if you have the observability disabled or expect it to fail.

jskupsik: OK to keep in - if the ERROR is annoying client can always suppress the service, especially since an app that gets this error wouldn't use the built-in ConnectionPoolMonitoringService and instead implement their own service to model their (different) jdbc pool.

jskupsik (approved):

PR #563, opened 2026-05-14
Clarified client app JDK 25 guidelines and pitfalls

jskupsik:


docs/application-structure.md
```diff
@@ -112,6 +112,41 @@ JVM arguments for `bootRun`, and extends `grails.build.info` with Hoist metadata
...
+We recommend **JDK 17** or **JDK 21** for client app builds. The published hoist-core JAR
+targets Java 17 bytecode and runs on any JDK 17+ runtime, so apps are not forced to track
+hoist-core's own build JDK.
+
+When building Docker images, use the matching xh-tomcat base image variant for your runtime
+JDK: `next-tc10-jdk17` or `next-tc10-jdk21`.
+
+#### JDK 25 on a Grails 7 / Gradle 8.x App
+
+Pointing Gradle directly at JDK 25 will fail:
```
lbwexler: if we could be just a bit more encouraging/positive in line 126 vs. 133 I am totally fine with this change

lbwexler: Let's not lead with the failure, but lead with the fact that it is fully supported, but requires running gradle on < JDK25

jskupsik: Fixed

PR #565, opened 2026-05-19
Fix recycled-facade IllegalStateException on servlet re-dispatches

lbwexler:
## Summary

Production logs show repeated `IllegalStateException: The request object has been recycled and is no longer associated with this facade` from `HoistFilter` → `TraceService.createSpan` → `hoistTags` → `IdentityService.getSession`.

Stack shows `ApplicationHttpRequest` in the call chain and `ErrorPageFilter` forwarding to `/error` — i.e. the failures happen on internal servlet re-dispatches (ERROR/FORWARD/INCLUDE/ASYNC), where the wrapped request points to a `RequestFacade` Tomcat has already recycled. `HoistFilter` was opening a SERVER span and touching headers/session unconditionally on every dispatch type.

## Fix

Gate `HoistFilter` on `DispatcherType.REQUEST`:
- REQUEST: full pipeline (trace context, SERVER span, auth) — unchanged.
- ERROR: route through `rethrowErrorDispatches` + `Utils.handleException` for consistent error rendering (preserves the "multipart size exceeded" use case the helper was added for). No SERVER span opened.
- FORWARD / INCLUDE / ASYNC: pass through to the chain. No tracing, no auth.

## Test plan

- [x] Build clean (verified locally).
- [x] Manual: trigger an unrouted URL (e.g. `/does-not-exist`), confirm no `IllegalStateException` in logs and that the response renders normally.
- [x] Manual: trigger multipart-size-exceeded, confirm error is rendered through hoist's exception pipeline.
- [x] Monitor production logs after deploy.

## Not in scope

The broader identity / async-safety refactor (PR #564) is parked — orthogonal to this fire.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## 41.0-SNAPSHOT - unreleased
+## 40.0.2 - 2026-05-19
 
 ### 🐞 Bug Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,11 +4,11 @@
 
 ### 🐞 Bug Fixes
 
-* Fixed `IllegalStateException: The request object has been recycled` thrown from
-  `HoistFilter` on internal servlet re-dispatches (ERROR/FORWARD/INCLUDE/ASYNC) — most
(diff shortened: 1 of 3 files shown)


lbwexler:
CR: Anselm

PR #570, opened 2026-07-06
Support nested view group renames + bulk view info updates for ViewManager

cnrudd:
Server-side support for the ViewManager nested-groups and bulk-editing work in xh/hoist-react#4479 (xh/hoist-react#4469). View groups are now slash-delimited paths supporting unlimited nesting (e.g. `Reports/Sales/Monthly`).

## Key changes (all non-breaking)

* New `ViewService.renameGroup` + `xhView/renameGroup` rename or re-parent a group and its entire subtree. Scope is an explicit argument - the global views, or those owned by the requesting user.
* New `JsonBlobService.renameGroup` provides the underlying support, rewriting `meta.group` across a type within a single owner namespace in one transaction.
* New `ViewService.bulkUpdateInfo` + `xhView/bulkUpdateInfo` apply the same metadata updates to multiple views in one call, reporting the token and reason for any failures.
* New `ExceptionHandler.isRoutine` classifies expected exceptions; `shouldLogDebug` now delegates to it.
* Fixed `updateInfo` and `create` failing when passed `isPinned` - the view type is now read from the view, not caller-supplied data.

Ships in v41 alongside hoist-react v87.


lbwexler:
Reviewed carefully.  Nice work guys.

PR #577, opened 2026-07-27
Add `BaseController.renderNdjson()` for streamed NDJSON responses

lbwexler:
- New `BaseController.renderNdjson()` streams an `Iterable`/`Iterator` to the client as NDJSON via Hoist's `JSONSerializer`, without materializing the full payload in memory. Pairs with `XH.fetchNdjson()` in hoist-react (unreleased).
- Flushes after the first element; setup failures still render a clean error response. A stream failing post-commit is terminated with a non-JSON line, so consumers detect truncation as a parse failure rather than loading a partial dataset as complete.
- `ExceptionHandler` now skips rendering to an already-committed response, avoiding corrupted output and duplicate log entries.
- Converted `parseRequestJSON`/`parseRequestJSONArray` to `@NamedVariant` — backward compatible.

### DIFF
--- a/grails-app/controllers/io/xh/hoist/BaseController.groovy
+++ b/grails-app/controllers/io/xh/hoist/BaseController.groovy
@@ -9,6 +9,8 @@ package io.xh.hoist
 
 import grails.async.Promise
 import groovy.transform.CompileStatic
+import groovy.transform.NamedParam
+import groovy.transform.NamedVariant
 import io.xh.hoist.cluster.ClusterService
 import io.xh.hoist.cluster.ClusterResult
 import io.xh.hoist.json.JSONParser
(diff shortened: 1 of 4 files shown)


grails-app/controllers/io/xh/hoist/BaseController.groovy
```diff
@@ -47,17 +50,71 @@ abstract class BaseController implements LogSupport, IdentitySupport {
...
+    /**
+     * Render a collection of objects as newline-delimited JSON (NDJSON), streaming each element
+     * to the client as it is serialized.
+     *
+     * Favor over {@link #renderJSON} for large row-oriented datasets — elements are written
+     * incrementally, so neither the full JSON string nor (for lazy sources) the full dataset is
+     * held in memory, and clients can parse rows as they arrive.
+     *
+     * The response is flushed (and committed) after the first element, so earlier failures still
+     * render a clean error response. A mid-stream failure cannot alter the committed status —
```
jskupsik, line 62: > A mid-stream failure cannot alter the committed status —
> instead the stream is terminated with a deliberately non-JSON line, so consumers fail
> parsing rather than mistaking the truncated (but otherwise well-formed) stream for a
> complete result.

This is *very* clever!

PR #584, opened 2026-08-01
Don't serialize JSONFormatCached's derived JSON into Hazelcast

amcclain:
`JSONFormatCached._cache` was not transient, so the derived JSON string was serialized along with the object into replicated `Cache`/`CachedValue`, `IMap`, and Topic structures, and into cross-instance call results — shipping and heap-retaining a redundant copy of data the object's own fields already carry.

Not a one-time cost at put time: Hoist uses `InMemoryFormat.OBJECT`, which serializes the *live* stored object on demand (remote read, partition migration, joiner sync), so a cache warmed any time after the put gets shipped, repeatedly.

### Changes, in priority order

1. **`transient` on `_cache`** — the actual fix. Each instance rebuilds lazily on first render.
2. **`volatile` on `_cache`** — concurrent readers reliably see a populated cache. Also required for (3).
3. **Double-checked locking in `getCachedJSON()`** — concurrent first-callers no longer each serialize the same object. Locks on `this`; a dedicated lock object would be null on instances Kryo deserializes via its Objenesis fallback, which skips constructors and field initializers.

(1) is the priority and stands on its own. (2) and (3) are negotiable in that order — happy to drop (3), or both, to keep the base class simpler. Benchmarked the locking locally against bulk serialization of large object arrays; overhead did not look problematic relative to the serialization work it guards.

### Heads-up for release

Changes how an existing *app-side* bug presents. For mutable subclasses — which the class docs warn against, but e.g. Toolbox's `Position` and `MarketPrice` are — one warm cache used to replicate, so every instance served identically stale JSON. Each instance now caches independently, so instances can serve divergent JSON for the same object.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,15 @@
   failures that are neither an exception nor an HTTP response. Avoids the need to synthesize a
   throwable - and its fabricated stack trace - purely to flag a span.
 
+### 🐞 Bug Fixes
+
+* `JSONFormatCached` now holds its cached JSON in a `transient` field, so it is no longer serialized
+  into Hazelcast structures (replicated `Cache`/`CachedValue`, `IMap`, Topics) or cross-instance
+  call results. Apps caching `JSONFormatCached` subclasses in replicated structures were shipping -
+  and holding in heap on every instance - a redundant copy of their own data. `getCachedJSON()` now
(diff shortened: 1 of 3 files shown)


lbwexler:
Tweaked the comments, which seemed like an overly confusing justifcation of standard java locking practices and keywords.

lbwexler:
Great change-- could effectively reduce a bunch of extra serialization work on server, I think!

PR #590, opened 2026-08-14
Add EntraIdService + DirectoryService for Entra ID directory group support

amcclain:
Adds an opt-in `EntraIdService` that queries Microsoft Entra ID (users, groups, transitive group memberships) via the Microsoft Graph API, and wires it into `DefaultRoleService` as an alternative to `LdapService` for directory-group role resolution. Targeted at hoist-core **v41.0.0** - see the included Changelog entry and `docs/upgrade-notes/v41-upgrade-notes.md` (🟢 LOW difficulty) for the app-facing summary.

## Key changes

- **`EntraIdService`** - app-only (client credentials) Graph queries via msal4j (token acquisition/caching) + Hoist's own `JSONClient` (REST). Reliability: bounded retry on 429/5xx, GUID validation, startup token warm-up, per-query caching, and strict/non-strict semantics that match the LDAP flow, so `DefaultRoleService`'s last-good-result protection covers both providers.
- **New configs** - `xhEntraIdConfig` (query settings + master switch), `xhEntraClientSecret`, and standalone `xhEntraTenantId` / `xhEntraClientId`. The tenant and client IDs stand alone so that other server-side subsystems can share them, and so that deployments can override them per environment via instance configs / env vars. Apps whose clients need them pre-auth (e.g. for OAuth login) relay them via their `AuthenticationService.getClientConfig()`.
- **`DirectoryService`** - narrow new interface implemented by `LdapService` and `EntraIdService`, modeling directory-group resolution for role management. `DefaultRoleService` auto-selects the enabled provider, with an optional `directoryGroupProvider` config key for the both-enabled case. Implementations throw cleanly when disabled - `DefaultRoleService` checks the enabled flag and degrades gracefully for the Admin Console.
- **`ErrorOr` typed results** (the one breaking change) - group resolution returns `Map<String, ErrorOr<Set<String>>>` in place of mixed `Set`-or-`String` values. Apps that override or call `doLoadUsersForDirectoryGroups` make a small mechanical update - see the upgrade notes. Wire formats are unchanged, so the client requires no changes.
- **Admin support** - new `DefaultRoleService.describeDirectoryGroups` / `searchDirectoryGroups` + matching `RoleAdminController` endpoints, so the Admin Console can show display names for assigned groups (Entra identifiers are opaque GUIDs) and search the directory by name - tokenized `$search` matching on Graph, substring matching across `sAMAccountName`/`cn` on LDAP. Client-side work tracked at xh/hoist-react#4572.
- **`xhLdapConfig.usernameAttribute`** - the LDAP person attribute mapped to the Hoist username (the `samaccountname` default preserves current behavior). Governs role-member resolution and is now also matched by `LdapService.lookupUser()` and `authenticate()`.
- **On-prem identity support** - `EntraUser` gains `onPremisesImmutableId` (with a decoded `onPremisesObjectGuid` accessor that handles the mixed-endian GUID byte order), `onPremisesSecurityIdentifier`, and `onPremisesSyncEnabled`, to support apps that join Entra accounts to on-prem AD identities. New `findUsers(field, value)` and `lookupUsers(ids)` methods support reverse and batch user lookups.
- **Hardening** - `LdapService` and `EntraIdService` are now `@CompileStatic`, parallel group lookups run in bounded batches of 25 as a backstop against fan-out storms, and both services refuse to authenticate with the `'none'` placeholder credentials - in particular, LDAP never binds with an unset query password, which could lock out the query account under an AD lockout policy.

## Docs

- New `docs/directory-services.md` feature doc, plus an accuracy and style refresh of `authentication.md` and `authorization.md` so the three read as a consistent triad.
- Finalized v41 Changelog entry and new `docs/upgrade-notes/v41-upgrade-notes.md`, both linked from the doc indexes and MCP registry.

Verified end-to-end against the XH Entra tenant via Toolbox: config hot-reload, token acquisition, and transitive member resolution through the Roles admin UI.


amcclain:
Companion Toolbox PR exercising this against the XH Entra ID tenant: xh/toolbox#888. Client-side admin UI work tracked at xh/hoist-react#4572.

amcclain:
I am now working on a companion hoist-react PR to add support to built-in role admin UI for searching groups and resolving (totally opaque) GUIDs to human readable names.

I will also validate that this impl serves the needs we have at the client that's prompting us to action this work now, including a need not directly related to role resolution where the Hoist server is being asked to relay some additional user facts we can query from MS Graph to a business API that needs them.

src/main/groovy/io/xh/hoist/directory/DirectoryService.groovy
```diff
@@ -0,0 +1,67 @@
...
+     * Resolve directory group identifiers to their member users, including members of any
+     * nested groups.
+     *
+     * <p>If strictMode is true, implementations must throw on any partial failure. Otherwise
+     * they log the failure and return an error description for the affected groups.
+     *
+     * @param groups - directory group identifiers, in the implementation-specific form
+     *      described by {@link #getDirectoryGroupsDescription}.
+     * @return Map of directory group identifier to either a Set of assigned usernames (on
+     *      success) or a String description of the lookup error (on failure).
```
lbwexler: We *could* look to move away from the Object representation here, and have this typed if we want, using Generics, e.g. 

Map<String, ErrorOr<Set<String>>

src/main/groovy/io/xh/hoist/role/provided/DefaultRoleService.groovy
```diff
@@ -174,56 +180,79 @@ class DefaultRoleService extends BaseRoleService {
...
+     * related Admin Console UI features.
+     *
+     * <p>Selected via the optional `directoryGroupProvider` key in `xhRoleModuleConfig` -
+     * `'ldap'`, `'entraId'`, or the default `'auto'`, which resolves to whichever single
+     * implementation is enabled. When both are enabled under auto, this method logs an ERROR
+     * and returns the LDAP implementation, so that an app experimenting with Entra ID does not
+     * silently switch its role memberships to the new source.
+     */
+    protected DirectoryService getDirectoryService() {
+        String provider = config.directoryGroupProvider ?: 'auto'
```
lbwexler, line 227: What does 'auto' buy us?  Seems confusing to me.  Does this have to do with bootstrapping?

amcclain:
Client-side counterpart now up: xh/hoist-react#4573 - Roles Admin display names + search-based group picker consuming the `directoryGroupsInfo` / `searchDirectoryGroups` endpoints added here. Degrades gracefully for apps on earlier hoist-core releases.

src/main/groovy/io/xh/hoist/role/provided/DefaultRoleService.groovy
```diff
@@ -174,56 +180,79 @@ class DefaultRoleService extends BaseRoleService {
...
+    /**
+     * Resolve display information for directory groups already assigned to roles, for the
+     * Admin Console UI. Returns a Map of group identifier to either a Map with a `displayName`
+     * key or a String error description.
+     */
+    Map<String, Object> describeDirectoryGroups(Set<String> groups) {
+        def svc = directoryService
+        if (!groups) return emptyMap()
+        if (!svc.enabled) {
+            return groups.collectEntries { [it, 'No enabled directory service in this application'] }
```
lbwexler: why not have all of the methods cleanly throw if no directory service?  This seems a little asymmetrical to me.

lbwexler (approved):
Approved -- main comments just have to do with the switching in DefaultRoleService.

src/main/groovy/io/xh/hoist/role/provided/DefaultRoleService.groovy
amcclain: Was looking to minimize amount of config entry changes we need to coordinate across all apps and all envs.  It did seem generally important to allow an app to have both ldap and entra services enabled, even though I don't think we'll have any apps in that state.  Given that, this flag seemed important to then direct which one to be used here.  But then I didn't necessarily want to e.g. default this flag to ldap and then to cutover you need to coordinate change to disable ldap and enable entra and change this flag without losing access to role resolution in the middle.

src/main/groovy/io/xh/hoist/directory/DirectoryService.groovy
amcclain: Yes you know I am in favor of the stronger typing and fewer maps-of-mixed-types returns - will look to work in this suggestion

src/main/groovy/io/xh/hoist/role/provided/DefaultRoleService.groovy
amcclain: Guards normalized so entra and ldap services all throw if the public APIs are hit when disabled, then adapted in the role layer to meet the existing contract with UI

src/main/groovy/io/xh/hoist/directory/DirectoryService.groovy
amcclain: Scoped this out - summary below so you can decide if you are up for the one breaking change involved.

**Sketch:** new `ErrorOr<T>` in `io.xh.hoist.util` with `of(value)` / `error(msg)` factories and a `success` flag. Signatures become `Map<String, ErrorOr<Set<String>>>` (load) and `Map<String, ErrorOr<Map>>` (describe). `ErrorOr` implements `JSONFormat` and serializes to the bare `error ?: value`.

**Pros**
- Kills the `instanceof Set` discrimination at all 3 consumer sites (`generateRoleAssignments`, `DefaultRoleAdminService.list`, `RoleAdminController`).
- Wire format is byte-identical, so zero hoist-react admin client changes.
- Self-documenting contract + IDE support, and a small general-purpose type we can reuse for future mixed-result APIs.
- Bonus: `_usersForDirectoryGroups` tightens from `Map<String, Object>` to `Map<String, Set<String>>`.

**Cons**
- **Breaking change to the protected `doLoadUsersForDirectoryGroups` override contract.** Apps that override it must wrap return values - a mechanical few-line change. Of our apps only Toolbox overrides it (mock directory groups); Jobsite / Veracity / DirectLend do not. An unmigrated override fails fast and loudly on first role refresh rather than corrupting anything.
- Groovy generics are unchecked at runtime, so this is documentation + tooling rather than enforcement.
- Changelog currently promises "apps that use LdapService today do not need to change" - would need a caveat for overriding apps, plus an upgrade note.

**Timing:** all of this lands in unreleased 41.0, so now is the moment. Fallback if you want to avoid the break: type only the `DirectoryService` interface and leave `doLoadUsersForDirectoryGroups` as `Map<String, Object>` - but then the two layers disagree, which is the asymmetry you flagged elsewhere. My vote is the full change. LMK and I will work it in.

lbwexler:
go for it

src/main/groovy/io/xh/hoist/directory/DirectoryService.groovy
amcclain: We've made this change - will be a new major release, I'll prep that shortly with upgrade notes to ensure apps/libraries make the change to the signature.

amcclain:
A number of other refinements added. Included a proactive review from within the requesting codebase of the private enterprise app that triggered these changes, which resulted in a few additional fields being added to our new `EntraUser` object and some more utilities added for finding and resolving user facts.

PR #592, opened 2026-08-20
Support a dedicated directory-reader app registration for EntraIdService

amcclain:
**Draft, on hold. Not ready to merge.**

Whether we want this at all depends on client signal. It is only worth taking if enterprise deployments go the shared directory-reader app registration route, rather than holding Graph application permissions on each application's own registration. Parking the work here so it is ready if that lands.

Stacked on #591. Base is `config-service-if-set-getters`, so the diff shows only the Entra work. #591 has to merge first.

### What it does

`EntraIdService` authenticates to Microsoft Graph as the application itself, using the shared `xhEntraClientId` and `xhEntraClientSecret`. That requires a client secret and admin-consented Graph application permissions on every application's own registration. This adds optional `xhEntraDirectoryClientId` and `xhEntraDirectoryClientSecret` configs to point the service at one shared registration instead.

- The client ID and secret resolve as a matched pair, so the service authenticates either wholly as the application or wholly as the directory reader. Pairing a directory secret with the application's own client ID is not reachable.
- `xhEntraClientId` keeps its existing meaning: the application's own registration, shared with client-side OAuth. Only this service follows the override.
- The tenant is deliberately not overridable. A directory in a second tenant would return group object IDs and usernames that do not line up with the tenant authenticating the application's own users, and cross-tenant app-only Graph access needs a multi-tenant registration consented into the target tenant.
- The resolved client ID is GUID-validated, named in every configuration error message, logged at startup alongside a `dedicatedDirectoryReader` flag, and reported in the service's Admin Console stats. Both configs are in `clearCachesConfigs`.

### Open question for review

Both configs are bootstrapped, so they appear in every app's config editor defaulted to `AppConfig.NONE`, including for apps that will never use them. That is deliberate: an un-bootstrapped config cannot be overridden by an instance config, because `AppConfig.overrideValue` only runs for an existing row, and per-environment override is the main reason to want these. The cost is two more rows in the config table.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,9 @@ detailed, step-by-step upgrade instructions with before/after code examples.
   `ErrorOr.of(usernames)` / `ErrorOr.error(message)` factories, and any code that calls
   `loadUsersForDirectoryGroups` and checks results with `instanceof Set` must read the new
   `ErrorOr.success` / `value` / `error` properties instead.
+* A blank instance config entry is now treated as unset, rather than resolving to an empty string.
+  Environment variables already behaved this way - file and YAML sources now match. Ignored keys are
+  logged at WARN during startup.
 
 ### 🎁 New Features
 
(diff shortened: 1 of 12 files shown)


amcclain:
Discussions with requesting client have leaned towards baking this into the primary app registration also used for the oauth flows, which would mean we do NOT need this additional flexibility at this time.  Noting this but keeping this open in draft form in case that guidance changes.  We have not yet rolled this out, so until we are done and in a steady operating state we can't be sure it won't change.

Issue #600, opened 2026-09-09
MCP symbol tools drop all per-argument documentation


amcclain:
## Summary

The MCP/CLI symbol tools drop all per-argument documentation. An agent calling `hoist-core-get-symbol` on a method sees the argument *names and types* but never their *descriptions*, even when the source documents every one. For a framework whose methods take large named-argument maps (`EmailService.sendEmail`, `TrackService.track`, `BaseService.createCache`), this is exactly the information an agent most needs.

Dropping it does not seem defensible. Either we surface it, or we document the limitation clearly so authors know that argument docs never reach agents.

## Evidence

`BaseService.parallelInit` documents both arguments:

```groovy
/**
 * Initialize a collection of BaseServices in parallel.
 * ...
 * @param services - BaseServices to initialize
 * @param timeout - maximum time to wait for each service to init (in ms).
 */
static void parallelInit(Collection<BaseService> services, Long timeout = 30 * SECONDS) {
```

What `hoist-core-get-symbol parallelInit` returns:

```
## Signature
static parallelInit(services: Collection<BaseService>, timeout: Long): void

## Documentation
Initialize a collection of BaseServices in parallel.
This is a blocking method.  Applications should make one or more calls to it to batch
initialize their services in an appropriate order.
```

Both `@param` lines are gone.

## Mechanism

Two independent causes, both in `mcp/src/main/groovy/io/xh/hoist/mcp/data/GroovyRegistry.groovy`:

1. **`@param` tags are parsed, then discarded.** `cleanDocComment` step 4 strips every line-anchored block tag:
   ```groovy
   // 4. Strip block tags (@param, @return, @throws, @since, ...) - line-anchored only,
   //    so mid-prose `{@link}` references aren't eaten.
   s = s.replaceAll(/(?m)^\s*@\w+\b.*$/, '')
   ```
   The content *is* present in the AST at this point - the raw Groovydoc reaching `cleanDocComment` still contains the `@param` lines. We throw it away.

2. **`ParameterInfo` has nowhere to put it.** It carries `name` and `type` only, and `extractGroovydoc` is called on the method node exclusively - never per-parameter.

## Why inline per-param comments are not a workaround

`TrackService.track` and `EmailService.sendEmail` currently document arguments with inline `/** */` comments inside the signature. Those are invisible to *every* consumer - Groovydoc, IntelliJ quick-doc, and the MCP - because Groovy's parser does not attach doc comments to parameter nodes at all.

[...]

PR #601, opened 2026-09-09
Convert EmailService.sendEmail to named parameters, harden address parsing

amcclain:
Converts `EmailService.sendEmail` to Groovy named parameters and fixes a set of address-parsing bugs found along the way.

### Named parameters

- `sendEmail` now declares its 13 arguments via `@NamedVariant` / `@NamedParam`, documented with `@param` Groovydoc.
- **No call-site changes required.** The generated `sendEmail(Map)` variant still accepts named args, a map literal, and a map held in a variable. All existing callers in hoist-core and Toolbox pass only valid keys.
- `@CompileStatic` callers now get compile-time errors for an unknown argument name, a wrong argument type, or a missing `to`. Dynamic callers get the same checks at runtime.
- ⚠️ An unrecognized argument name now throws, where it was previously ignored. The check runs before `sendEmail`, so `throwError: false` does not suppress it, and `AssertionError` is an `Error`, so `catch (Exception)` around the call will not catch it.

### Address-parsing fixes

- Blank entries are discarded, so a blank config no longer expands into an invalid address such as `@example.com`. A blank `xhEmailOverride` was the worst case - that address is truthy, so it redirected all mail to it.
- The `xhEmailDefaultDomain` suffix is skipped when that config is itself blank or `none`, rather than appending a bare `@`.
- `parseAddresses` reads `none` in any letter case.
- `sendEmail` throws a clear error when it cannot resolve a sender address.

### Also

- `EmailService` is now `@CompileStatic`. This required binding the mail builder explicitly inside the `sendMail` closure, because the named parameters shadow its same-named DSL methods, and dispatching `attach` on the `contentSource` type.
- `docs/email.md` and the CHANGELOG updated.

Related: #600 tracks the MCP symbol tools dropping per-argument documentation, which is why the `@param` tags here are not yet visible to agents.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,20 @@
 
 ## 42.0-SNAPSHOT - unreleased
 
+### 🐞 Bug Fixes
+
+* Hardened address parsing in `EmailService`. A blank entry no longer expands into an invalid
+  address such as `@example.com`, which a blank `xhEmailOverride` config applied to all mail. The
+  service also skips a blank or `none` `xhEmailDefaultDomain`, reads `none` in any letter case, and
+  throws a clear error when no sender address resolves.
+
+### ⚙️ Technical
+
+* `EmailService.sendEmail` now declares its arguments as named parameters. Existing calls need no
+  changes.
+    * ⚠️ A call that passes an unrecognized argument name now fails with an `AssertionError`. The
+      check runs before `sendEmail`, so the `throwError: false` default does not suppress it.
+
 ## 41.0.0 - 2026-08-25
 
 ### 💥 Breaking Changes (upgrade difficulty: 🟢 LOW - most apps require no changes)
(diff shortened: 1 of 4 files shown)


amcclain:
⚠️ **Not yet tested against a live mail server.** Nothing here has actually sent an email.

What has been verified:

- `./gradlew clean assemble` and `./gradlew groovydoc` pass.
- Bytecode inspection confirms all 11 `MailMessageBuilder` calls inside the `sendMail` closure still dispatch statically after the `@CompileStatic` conversion.
- Ad-hoc scripts (throwaway, not checked in - hoist-core has no test suite) covered 38 assertions against the compiled class: named-arg mapping and every default, the three legacy call styles, unknown-key and missing-`to` rejection, address parsing, and `filterAddresses`.
- Compile-time checking confirmed by compiling deliberately bad call sites under `@CompileStatic`.

What that leaves unexercised - everything past the early-outs in `sendEmail`, which needs a Grails/Spring context:

- The actual send, including the `sendMail` closure and `MailMessageBuilder` interaction.
- Attachment type dispatch for `byte[]` / `File` / `InputStreamSource`.
- Filter and override application, subject enhancement, and logging.

Attachments and the `xhEmailOverride` path are worth a manual pass before this comes out of draft, since both were touched and neither is covered.

PR #602, opened 2026-09-10
Route all directory group operations through getDirectoryService

lbwexler:
`roleAdmin/directoryGroupsInfo` reported `No enabled directory service in this application` for every assigned group in any app that resolves directory groups from its own source (Toolbox, via its mock, is the case in hand). `DefaultRoleService.describeDirectoryGroups` and `searchDirectoryGroups` went straight to the selected `DirectoryService`, while only membership resolution had an app-level hook - `doLoadUsersForDirectoryGroups`. That asymmetry was the bug.

`DirectoryService` is already the narrow contract for exactly these operations, and `getDirectoryService()` is already protected, so the fix removes a seam rather than adding two.

Key changes:

- Removed `DefaultRoleService.doLoadUsersForDirectoryGroups` and the `final loadUsersForDirectoryGroups` wrapper that only called it. `loadUsersForDirectoryGroups` is now a plain public method with that same body, symmetric with the other two.
- Documented `getDirectoryService()` as the single override seam for group resolution, display names, and group search alike.
- Updated `docs/authorization.md` and `docs/directory-services.md` to point at the new seam.
- Changelog: a Breaking Changes bullet for the removal, plus a Bug Fixes bullet for the bogus errors.

Toolbox companion PR replaces its `doLoadUsersForDirectoryGroups` override with a `MockDirectoryService`: https://github.com/xh/toolbox/pull/902

Two notes for review:

- I opened the `42.0-SNAPSHOT` Breaking Changes section and rated it `MEDIUM - Grails 8 / Groovy 5` to match the in-flight v42 upgrade notes. The Grails 8 bullets themselves are not in the changelog yet.
- `docs/upgrade-notes/v42-upgrade-notes.md` is deliberately not part of this PR - the v42 notes get generated when the time is right. The changelog's link to that file therefore dangles until they land, and the upgrade step for this change ("Replace a `doLoadUsersForDirectoryGroups` override") goes in then.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,22 @@
 
 ## 42.0-SNAPSHOT - unreleased
 
+### 💥 Breaking Changes (upgrade difficulty: 🟠 Minor)
+
+See [`docs/upgrade-notes/v42-upgrade-notes.md`](docs/upgrade-notes/v42-upgrade-notes.md) for
+detailed, step-by-step upgrade instructions with before and after code examples.
+
+* Removed `DefaultRoleService.doLoadUsersForDirectoryGroups`. Apps that resolve directory groups
+  from a custom source now override `getDirectoryService` to return their own `DirectoryService`
+  implementation. Apps that do not override the removed method require no changes.
+
+### 🐞 Bug Fixes
+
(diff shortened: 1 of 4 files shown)


lbwexler:
CR: Colin

PR #604, opened 2026-09-15
Cache the parsed Entra certificate instead of re-parsing per admin stats read

lbwexler:
Builds on #596 - targets `entra-cert-auth`, not `develop`.

The service was unpacking the certificate bundle every time something asked it for admin stats. All it wanted was the expiry date and thumbprint to display, but getting those meant re-reading the config, base64-decoding it, and opening the PKCS#12 keystore again from scratch. That's not free, and it happens on every instance in the cluster each time a cluster report runs.

So now the bundle gets opened once - when the service first builds its credential, which it was already doing - and the certificate is kept next to it. The admin stats just read what's already sitting there. Clearing caches throws away the credential along with everything else, so swapping in a new certificate still takes effect without a restart, and it reloads eagerly the same way `init` does at startup.

Three things fell out of that:

* The thumbprint is now cheap enough to include in the cross-instance comparison, so a cluster report will flag an instance that's somehow running a different certificate than its peers.
* The code that picks the certificate out of the bundle is only reachable after MSAL has already validated it, which closes a small gap where our code would have quietly accepted a bundle the docs say is rejected.
* `getAdminStats` is back to a plain map literal like every other service - `ClusterObjectInfo` already catches and logs anything it throws, and surfaces it as a proper `error` field.

One deliberate wart: the bundle is opened twice at startup, once by MSAL to build the credential and once by us to read the certificate back out for display. That keeps the private key out of Hoist's hands entirely and leaves MSAL as the one enforcing what a valid bundle looks like. The single-load alternative means reimplementing `ClientCertificate.create` - including its null-chain fallback - and taking over the single-key-entry check your troubleshooting table quotes MSAL's error message for. One extra keystore load per credential lifetime seemed the better trade.

Docs: consolidated the duplicated "upload the public certificate" instruction under its own heading, and replaced the macOS-only `base64 -i` (GNU reads it as `--ignore-garbage` and hangs on stdin) with a single pipeline matching the ConfigSpec note.

No behavior change to token acquisition - the MSAL credential is built from the same call with the same inputs.


src/main/groovy/io/xh/hoist/entra/EntraClientCredentials.groovy
```diff
@@ -36,28 +48,54 @@ class EntraClientCredentials {
...
      */
-    static X509Certificate parseCertificate(String pfxBase64, String pfxPassword) {
-        char[] password = (pfxPassword ?: '').toCharArray()
+    Map getCertificateStats() {
+        if (!certificate) return null
+        return [
+            subject: certificate.subjectX500Principal.name,
+            // SHA-1 hex, matching the "Thumbprint" column shown in the Azure portal.
+            thumbprint: MessageDigest.getInstance('SHA-1').digest(certificate.encoded).encodeHex().toString().toUpperCase(),
+            notAfter: certificate.notAfter
```
amcclain, line 86: I am totally fine sticking with `notAfter` but if this is a timestamp would it be beneficial to rename if we could get our admin client auto date formatting activated?

amcclain (approved):
Thank you for taking a thorough look over this and improving!

src/main/groovy/io/xh/hoist/entra/EntraClientCredentials.groovy
lbwexler: great catch, done

